PR #10, opened 2024-12-23
chore: improve demo upload data to handle fatal errors

DominicGBauer:
## Description
* Improve demo `uploadData` function to handle fatal postgres errors.
* Fix link error in docs

## Testing
I added RLS to `lists` table in Supabase and then added new a list. Initially the error was not handled after the change the error is now handled and the queue is not blocked.

### DIFF
--- a/Demo/PowerSyncExample/PowerSync/SupabaseConnector.swift
+++ b/Demo/PowerSyncExample/PowerSync/SupabaseConnector.swift
@@ -4,11 +4,45 @@ import Supabase
 import PowerSync
 import AnyCodable
 
+private struct PostgresFatalCodes {
+    /// Postgres Response codes that we cannot recover from by retrying.
+    static let fatalResponseCodes: [String] = [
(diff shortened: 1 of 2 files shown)


benitav (approved):

Demo/PowerSyncExample/PowerSync/SupabaseConnector.swift
```diff
@@ -4,11 +4,45 @@ import Supabase
 import PowerSync
 import AnyCodable
 
+private struct PostgresFatalCodes {
+    /// Postgres Response codes that we cannot recover from by retrying.
+    static let fatalResponseCodes: [String] = [
```
HeinrichvonStein, line 9: Can we recover from deadlocks or serialization failures?

DominicGBauer: Not sure

PR #12, opened 2025-01-09
chore: update powersync-sqlite-core

DominicGBauer:
## Description
Update `powersync-sqlite-core` package

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@ let package = Package(
             targets: ["PowerSync"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/powersync-ja/powersync-kotlin.git", exact: "1.0.0-BETA11.0"),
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -18,7 +18,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/powersync-ja/powersync-kotlin.git", exact: "1.0.0-BETA11.0"),
```
rkistner, line 20: Should this also be updated?

rkistner (approved):

Package.swift
DominicGBauer: Yeah I've updated it now

rkistner (approved):

DominicGBauer:
Merging with known transaction issue that is causing the test to fail

PR #15, opened 2025-01-30
feat: update to latest kotlin and apply breaking changes to demo

DominicGBauer:
## Description
Update underlying `powersync-kotlin` SPM package which introduces a breaking change to transactions where the callback has become synchronous.

Also updated the demo implementation based on this breaking change. 

## Testing
Ran the demo app and ran Crud operations from Supabase and the device.


stevensJourney (approved):
Looks good to me

PR #24, opened 2025-02-20
WIP feat: update package and improve demo

DominicGBauer:


stevensJourney:
I believe most of the changes here are no longer required after the watched query stability improvements.

PR #26, opened 2025-03-05
Watched Query Fixes

stevensJourney:
# Overview

This applies the latest watched query updates from https://github.com/powersync-ja/powersync-kotlin/pull/136.

The current Query protocol is a bit verbose due to handling optional function parameters. Watched queries now support an optional `throttleMs` parameter. Adding this new param to the current structure would require a function definition for every possible permutation of optional arguments. A new struct based approach has been added for the `watch` method, this handles optional arguments better.

### DIFF
--- a/Sources/PowerSync/Kotlin/KotlinPowerSyncDatabaseImpl.swift
+++ b/Sources/PowerSync/Kotlin/KotlinPowerSyncDatabaseImpl.swift
@@ -183,6 +183,14 @@ final class KotlinPowerSyncDatabaseImpl: PowerSyncDatabaseProtocol {
                 Task {
                     do {
                         var mapperError: Error?
+                        // HACK!
+                        // SKIEE doesn't support custom exceptions in Flows
+                        // Exceptions which occur in the Flow itself cause runtime crashes.
+                        // The most probable crash would be the internal EXPLAIN statement.
+                        // This attempts to EXPLAIN the query before passing it to Kotlin
+                        // We could introduce an onChange API in Kotlin which we use to implement watches here.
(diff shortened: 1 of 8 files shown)


Sources/PowerSync/Kotlin/KotlinPowerSyncDatabaseImpl.swift
```diff
@@ -165,38 +165,42 @@ final class KotlinPowerSyncDatabaseImpl: PowerSyncDatabaseProtocol {
...
+    }
+
+    func watch<RowType>(
+        options: WatchOptions<RowType>
     ) throws -> AsyncThrowingStream<[RowType], Error> {
         AsyncThrowingStream { continuation in
             Task {
                 do {
                     var mapperError: Error?
+                    // HACK!
```
simolus3, line 186: My preferred solution for this would be to add a utility wrapping errors to our SDK, e.g. in `iosMain` of the `core` package:

```Kotlin
package com.powersync

import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.catch
import kotlinx.coroutines.flow.map

public val <T>  Flow<T>.asResult: Flow<Result<T>> get() = this
    .map(Result.Companion::success)
    .catch { error -> emit(Result.failure(error)) }
```

Then, we could use that as a catchall in the Swift SDK when we're listening on Flows.

simolus3 (approved):
The Swift API changes look good to me. We can release the workaround for now, but I think we should look into other approaches that will be able to catch all exceptions.

simolus3 (approved):

PR #34, opened 2025-04-07
Configurable Logging

stevensJourney:
# Overview

This adds the ability to specify a `logger` when creating a PowerSync database. The minimum log severity can be configured when creating a `DefaultLogger`.

Loggers also have the ability to specify custom `LogWriter`s which can transport logs to custom sinks.

The implementation here declares a Swift logging interface and maps this to a Kotlin Kermit logger which is used internally.

### DIFF
--- a/Sources/PowerSync/Kotlin/DatabaseLogger.swift
+++ b/Sources/PowerSync/Kotlin/DatabaseLogger.swift
@@ -1,0 +1,135 @@
+import PowerSyncKotlin
+
+/// Maps a Kermit `Severity` level to a local `LogSeverity`.
+///
+/// - Parameter severity: The Kermit log severity value from Kotlin.
+/// - Returns: The corresponding `LogSeverity` used in Swift.
+private func mapKermitSeverity(_ severity: PowerSyncKotlin.Kermit_coreSeverity) -> LogSeverity {
+    switch severity {
+        case PowerSyncKotlin.Kermit_coreSeverity.verbose:
+            return LogSeverity.debug
+        case PowerSyncKotlin.Kermit_coreSeverity.debug:
+            return LogSeverity.debug
+        case PowerSyncKotlin.Kermit_coreSeverity.info:
+            return LogSeverity.info
+        case PowerSyncKotlin.Kermit_coreSeverity.warn:
+            return LogSeverity.warning
+        case PowerSyncKotlin.Kermit_coreSeverity.error:
+            return LogSeverity.error
+        case PowerSyncKotlin.Kermit_coreSeverity.assert:
+            return LogSeverity.fault
+    }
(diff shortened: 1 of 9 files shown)


simolus3 (commented):
I'm in favor of making the loggers customizable, but I'm wondering why we're adding our own interfaces here instead of e.g. using https://github.com/apple/swift-log? We have to write a bridge from Kermit to _something_ either way, so why not something that has an existing ecosystem with plenty of writer implementations around it?

Sources/PowerSync/Kotlin/DatabaseLogger.swift
```diff
@@ -0,0 +1,106 @@
...
+    init(_ logger: any LoggerProtocol) {
+        self.logger = logger
+        // Set to the lowest severity. The provided logger should filter by severity
+        kLogger.mutableConfig.setMinSeverity(Kermit_coreSeverity.verbose)
+        kLogger.mutableConfig.setLogWriterList(
+            [KermitLogWriterAdapter(logger: logger)]
+        )
+    }
+    
+    /// Sets the minimum severity level that will be logged.
```
benitav, line 64: Is it quite standard what the levels are are in what order they are? If not we should maybe list the ordered options here.

simolus3 (dismissed):
I see 👍 In my humble opinion, using `swift-log` would make things easier for our users out of the box and reduce the code we have to write around defining logger APIs and printing to `OSLog`. But I understand the desire to not introduce dependencies either, and this approach looks good to me as well.

Sources/PowerSync/Kotlin/DatabaseLogger.swift
stevensJourney: The options are taken from the standard Swift OSLogger interface. They are listed in the `LogSeverity` enum.

```Swift
public enum LogSeverity: Int, CaseIterable {
    /// Detailed information typically used for debugging.
    case debug = 0

    /// Informational messages that highlight the progress of the application.
    case info = 1

    /// Potentially harmful situations that are not necessarily errors.
    case warning = 2

    /// Error events that might still allow the application to continue running.
    case error = 3

    /// Serious errors indicating critical failures, often unrecoverable.
    case fault = 4

    /// Map severity to its string representation
    public var stringValue: String {
        switch self {
        case .debug: return "DEBUG"
        case .info: return "INFO"
        case .warning: return "WARNING"
        case .error: return "ERROR"
        case .fault: return "FAULT"
        }
    }

    /// Convert Int to String representation
    public static func string(from intValue: Int) -> String? {
        return LogSeverity(rawValue: intValue)?.stringValue
    }
}
```

simolus3 (approved):
LGTM 👍

PR #41, opened 2025-04-24
API Reference 

stevensJourney:
# Overview

This adds a basic automatically generated API reference for the Swift SDK.

Docs are generated with XCode's docbuild command.

Docs are deployed  at https://powersync-ja.github.io/powersync-swift

Docs and other SwiftDoc comments will be updated as future work. 

image

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,20 @@
+name: Deploy Docs
+
+on:
+  push:
+    branches:
+      - main
+      - api-ref
+
+jobs:
+  build:
+    name: Build and test
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build Docs
+        run: |
+          xcodebuild docbuild -scheme PowerSync -destination 'platform=macOS' -derivedDataPath ./DerivedData
+          xcrun docc process-archive transform-for-static-hosting \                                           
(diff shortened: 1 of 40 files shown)


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,46 @@
+name: Deploy Docs
+
+on:
+  push:
```
stevensJourney, line 4: TODO. Need to limit this to main before merging

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,54 @@
...
+        id: deployment
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: docs-site
+
+  # Deployment job
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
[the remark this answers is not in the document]
stevensJourney: It's interesting how the workflow worked without this. Updated the implementation.

.github/workflows/docs.yaml
rkistner: Does this still need to be done?

rkistner (dismissed):

.github/workflows/docs.yaml
stevensJourney: This was added temporarily in order to have the served Github page be available during the review process. I've now limited the branch to `main`

rkistner (approved):

PR #44, opened 2025-05-02
Enable macOS support for demo

simolus3:
Enable the macOS target for the demo project. The app is not particularly pretty, but it's a native macOS app:

macOS demo

I also had to disable the code related to attachments on macOS because most of the APIs related to the camera or images seem to be different.


stevensJourney (approved):
Awesome work!

PR #45, opened 2025-05-05
Add sync progress APIs

simolus3:
This ports the sync progress APIs over from Kotlin (using an identical API mirrored with Swift protocols). It also adds an example to the demo app which now shows a progress bar for the first sync.


stevensJourney (approved):
Looks good to me.

PR #56, opened 2025-07-08
Update README.md - remove Beta

cahofmeyr:
No longer in Beta: https://releases.powersync.com/announcements/powersync-swift-sdk-v1-0-0

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,6 @@ _[PowerSync](https://www.powersync.com) is a sync engine for building local-firs
 
 This is the PowerSync SDK for Swift clients. The SDK reference is available [here](https://docs.powersync.com/client-sdk-references/swift).
 
-## Beta Release
-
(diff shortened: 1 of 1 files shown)


cahofmeyr:
> In the `Installation` section, can we replace the exact dependency with a semver range suggestion?
> 
> ```swift
> dependencies: [
>         ...
>         .package(url: "https://github.com/powersync-ja/powersync-swift", from: "<version>")
> ],
> ```
> 
> Now that we're not recommending pinning an exact version anymore, this feels like a better default.

@stevensJourney what are your thoughts?

stevensJourney:
Thanks for catching and removing the BETA notice 😅 .

For the dependencies. I think we should be able to update the READE _example_ to 

```swift
dependencies: [
        ...
        .package(url: "https://github.com/powersync-ja/powersync-swift", from: "1.0.0")
],
```

My understanding is that from: "1.0.0" follows the SemVer range equivalent to ^1.0.0, which means it will install the latest compatible SDK for new users and automatically upgrade as we release new versions. Users are also free to pin the version if they prefer.

I can commit those changes to this branch in order to get this over the line.

cahofmeyr:
@simolus3 @stevensJourney I see the README still mentions KMMBridge; should that also be removed? 
- https://github.com/powersync-ja/powersync-swift?tab=readme-ov-file#underlying-kotlin-dependency
- https://github.com/powersync-ja/powersync-docs/pull/193

PR #63, opened 2025-07-31
Add network logging support

stevensJourney:
# Overview

This adds the ability to log PowerSync network requests. This is the Swift port of https://github.com/powersync-ja/powersync-kotlin/pull/229.

This also removes the experimental `ConnectionMethod` options which have recently been removed from the Kotlin SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.4.0
+## 1.4.0 (unreleased)
 
 * Added the ability to log PowerSync sync network requests.
 
@@ -18,6 +18,7 @@ try await database.connect(
         )
     )
 )
+
 ```
 ## 1.3.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## 1.4.0
 
-* Added the ability to log PowerSync network requests.
+* Added the ability to log PowerSync sync network requests.
+
 ```swift
-struct InlineLogger: NetworkLogger {
+struct InlineLogger: SyncRequestLogger {
     func log(_ message: String) {
         print("Network: \(message)")
     }
@@ -14,7 +15,7 @@ try await db.connect(
                 connector: connector,
                 options: ConnectOptions(
                     clientConfiguration: SyncClientConfiguration(
-                        networkLogger: NetworkLoggerConfig(
+                        requestLogger: SyncRequestLoggerConfiguration(
                             logLevel: .headers,
                             logger: InlineLogger()
                         )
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
+## 1.4.0
+
+* Added the ability to log PowerSync sync network requests.
+
+```swift
+struct InlineLogger: SyncRequestLogger {
```
simolus3, line 8: Is there a reason we're not re-using `LoggerProtocol` for this? Could the `SyncRequestLoggerConfiguration` take a `LoggerProtocol`, a severity (probably defaulting to something verbose) and the log level and then post messages there?

stevensJourney: I initially considered reusing `LoggerProtocol`, but opted for the current `SyncRequestLogger`.

The PowerSync network logger is focused specifically on HTTP request/response data, and uses a custom "log-level" `SyncRequestLogLevel` which controls what type of network data is emitted (headers, body, etc.), rather than traditional severity levels like `.debug`, `.info`, etc.

If we used `LoggerProtocol`, users would have to provide: a `logger` instance, a `SyncRequestLogLevel`  and a corresponding `LoggerLevel` to use for all network messages. That felt slightly more complex and potentially confusing. With `SyncRequestLogger`, we give devs full control to decide how to map messages to their own logging system or levels inside `log(_:)`.

There might not be a strong use case for this, but `SyncRequestLogger` also makes it slightly easier to process network logs without the use of a traditional logger.

simolus3: Fair. I wonder if there should be another constructor on `SyncRequestLoggerConfiguration` that takes the logger, `SyncRequestLogLevel` and `LoggerLevel`. That way users who already have a logger don't need to implement another interface, and those who want them separate can use `SyncRequestLogger`?

stevensJourney: 👍 I like that idea. I've implemented your suggestion of an additional constructor. I've also removed `SyncRequestLogger` for a single closure based approach.

Sources/PowerSync/Protocol/SyncRequestLogger.swift
```diff
@@ -0,0 +1,83 @@
...
+    case none
+}
+
+/// Configuration for PowerSync HTTP request logging.
+///
+/// This configuration is set once during initialization and used throughout
+/// the PowerSync session. The `requestLevel` determines which network events
+/// are logged.
+///
+/// - Note: The request levell cannot be changed after initialization. A new call to `PowerSyncDatabase.connect` is required to change the level.
```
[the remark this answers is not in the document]
stevensJourney: Fixed

simolus3 (approved):
Looks good to me!

PR #69, opened 2025-08-21
[Internal] Improve Lock Error handling

stevensJourney:
# Overview

It's currently tricky to throw Errors from Swift closure functions which are provided to the Kotlin SDK.

We currently achieve exception propagation for the `readLock`, `readTransaction`, `writeLock` and `writeTransaction` methods by implementing `TransactionCallback` and `LockCallback` wrappers on the Swift side. These wrappers:

- Run the provided Swift closure and catch any thrown exceptions
- Return the exception as the result of the wrapped closure as a PowerSyncException instance
- The Kotlin SDK then detects this exception and throws the exception on the Kotlin side

SKIEE has known limitations with generics. Which causes some rather strange requirements for the `TransactionCallback` and `LockCallback` implementations.

The `execute` method of these implementations should be able to return `Any?` (a nullable response). But this is not possible due to Objective C limitations:
```
Throwing method cannot be an implementation of an '@objc' requirement because it returns a value of optional type 'Any?'; 'nil' indicates failure to Objective-CSourceKit
```

This causes us to define the `execute` method as returning `Any` which works, but prints a warning on every compile.

```
func execute(transaction: PowerSyncKotlin.PowerSyncTransaction) throws -> Any {
          - warning: result of 'execute(transaction:)' has different optionality than expected by protocol 'ThrowableTransactionCallback'
52          do {
53              return try callback(
```

Furthermore. When enabling the Swift 6 toolchain while making Swift 6 declaration changes. The compiler seems to crash when processing the `TransactionCallback.swift` file. 

This PR removes the need for `TransactionCallback` and `LockCallback` in the Swift SDK, by making use of the changes in Kotlin PR


simolus3 (approved):
This looks good to me, I also like that this lets us remove Swift-specific logic from the `:core` Kotlin project.

simolus3 (approved):

PR #74, opened 2025-09-02
Fix broken local uri's when relaunching the app

jonkan:
Hey! I asked about this in the [discord](https://discord.com/channels/1138230179878154300/1194710422960472175/1412156165915414658) as well.

Since simulators update the path to the app's sandbox each time the app is run I'm hitting this issue where the local uri's stored in the database break the next time I run the app.

What I've observed:

State A: clean slate, not signed in, no sync enabled (yet).
1. Create an attachment. Everything looks good.
2. Restart the app. Local uri is broken, image fails to load from the local uri.
3. Sign in, sync is enabled.
4. The attachment gets the state .archived in the verifyAttachments, because exists is false.
5. The attachment then gets state .queuedDownload in the processWatchedAttachments, because hasSynced is false and it has a localUri.
6. It tries to download the attachment and gets 404 (expected, because it hasn't been uploaded yet).
7. Retries the downloading every 1 sec indefinitely (if no errorHandler, which is the default).

State B, already signed in:
1. Create an attachment. 
2. Attachment is uploaded as expected.
3. Restart the app.
4. Local uri is broken.
5. Attachment is download.
6. Everything looks good again.

I've created a partial fix in this PR, by updating the local uri's in the verifyAttachments. The issue that remains is that verifyAttachments isn't run until sync is enabled. I.e. I'm still hitting the bug in State A until step 2.

Can we run verify earlier? Other input/ideas?

Also, this should only be an issue when running on a Simulator. Not for real devices.

### DIFF
--- a/Demo/PowerSyncExample.xcodeproj/project.pbxproj
+++ b/Demo/PowerSyncExample.xcodeproj/project.pbxproj
@@ -3,10 +3,14 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 60;
+	objectVersion = 56;
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0B29DBE92E686D6000D60A06 /* FtsSetup.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B29DBE82E686D5A00D60A06 /* FtsSetup.swift */; };
+		0B29DBEB2E68876500D60A06 /* SearchResultItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B29DBEA2E68875C00D60A06 /* SearchResultItem.swift */; };
+		0B29DBED2E68887A00D60A06 /* SearchScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B29DBEC2E68887700D60A06 /* SearchScreen.swift */; };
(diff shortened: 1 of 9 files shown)


Sources/PowerSync/attachments/AttachmentQueue.swift
```diff
@@ -463,7 +463,15 @@ public actor AttachmentQueue: AttachmentQueueProtocol {
...
-            if attachment.state == AttachmentState.queuedUpload {
+            let newLocalUri = await getLocalUri(attachment.filename)
+            let newExists = try await localStorage.fileExists(filePath: newLocalUri)
+            if newExists {
+                // The file exists but the localUri is broken, lets update it.
+                // E.g. this happens in simulators that change the path to the app's sandbox.
+                updates.append(attachment.with(
+                    localUri: newLocalUri
+                ))
+            } else if attachment.state == AttachmentState.queuedUpload || attachment.state == AttachmentState.archived {
```
stevensJourney, line 474: I like the approach here for checking and updating the URI.

stevensJourney (commented):
Thanks! The fix here looks good to me. 

I think it should be fine to run `verifyAttachments` earlier (in addition to running it when syncing starts).

The simplest solution might be to expose a public method like `verifyAttachments` which users can call before `startSync`.

stevensJourney (approved):
@jonkan Thanks. This looks good to me. I'll add some minor items like a changelog entry and potentially some of the items mentioned above in a follow up PR.

PR #75, opened 2025-09-03
Additional Strict Concurrency Declarations

stevensJourney:
# Overview

https://github.com/powersync-ja/powersync-swift/pull/70 added initial support for Swift 6 Strict Concurrency checks by annotating/extending protocols with `Sendable`.

The initial work missed some of our API protocols and definitions. This PR declares remaining items as Sendable. 

The example app has also been updated to use Swift 6 with strict concurrency checking enabled.

### DIFF
--- a/Sources/PowerSync/Kotlin/sync/KotlinSyncStatusData.swift
+++ b/Sources/PowerSync/Kotlin/sync/KotlinSyncStatusData.swift
@@ -103,11 +103,17 @@ extension KotlinProgressWithOperationsProtocol {
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/PowerSync/Protocol/Schema/Table.swift
```diff
@@ -93,7 +93,7 @@ public struct Table: TableProtocol {
         viewNameOverride ?? name
     }
 
-    internal var internalName: String {
+    var internalName: String {
```
simolus3, line 96: Out of interest, why did this have to change?

simolus3 (approved):
Looks good to me 👍

PR #79, opened 2025-09-15
Update Kotlin and core extension

simolus3:
Release notes for the core extension: https://github.com/powersync-ja/powersync-sqlite-core/releases/tag/v0.4.6

This also prepares the adoption of a Kotlin SDK update to fix XCode 26 compatibility (https://github.com/powersync-ja/powersync-kotlin/pull/260).
It also adds the `getCrudTransactions` API we've recently added to our other SDKs.

### DIFF
--- a/Demo/PowerSyncExample/PowerSync/FtsSetup.swift
+++ b/Demo/PowerSyncExample/PowerSync/FtsSetup.swift
@@ -159,10 +159,11 @@ func configureFts(db: PowerSyncDatabaseProtocol, schema: Schema) async throws {
     // --- Execute all generated SQL statements ---
 
     if !allSqlStatements.isEmpty {
+        let resultingStatements: [String] = allSqlStatements
         do {
             print("[FTS] Executing \(allSqlStatements.count) SQL statements in a transaction...")
             _ = try await db.writeTransaction { transaction in
-                for sql in allSqlStatements {
+                for sql in resultingStatements {
                     print("[FTS] Executing SQL:\n\(sql)")
                     _ = try transaction.execute(sql: sql, parameters: [])
                 }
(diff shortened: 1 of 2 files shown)


stevensJourney (approved):

simolus3:
Sounds good, I'll open a Kotlin PR for this 👍

Sources/PowerSync/Protocol/db/CrudTransaction.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import PowerSyncKotlin
```
stevensJourney, line 2: The implementation here looks great to me. 
While this is not strictly required. I like to keep `PowerSyncKotlin` out of the files which define the Swift Protocols. Previously mixing these has made it very easy to leak PowerSyncKotlin into our public APIs - we aren't leaking anything here, but to make it a little easier to prevent in future it would be nice if we could perhaps declare `CrudTransactions` as a Protocol or typealias here and link it to the Kotlin implementation in a file in the `Kotlin` folder.

Demo/PowerSyncExample/PowerSync/FtsSetup.swift
```diff
@@ -159,10 +159,11 @@ func configureFts(db: PowerSyncDatabaseProtocol, schema: Schema) async throws {
     // --- Execute all generated SQL statements ---
 
     if !allSqlStatements.isEmpty {
+        let resultingStatements: [String] = allSqlStatements
         do {
             print("[FTS] Executing \(allSqlStatements.count) SQL statements in a transaction...")
             _ = try await db.writeTransaction { transaction in
-                for sql in allSqlStatements {
+                for sql in resultingStatements {
```
simolus3, line 166: Had to make this immutable to stop Swift complaining about this potentially being used concurrently.

stevensJourney (approved):

PR #82, opened 2025-10-17
Encryption support

simolus3:
With the latest CSQLite and Kotlin SDK release, we can finally support encryption in the Swift SDK as well. On the SDK, the only change is that we add a parameter controlling which statements to run when databases are opened. This allows running a `pragma key = ` to set encryption keys.

Actually using encryption is a bit trickier:

- The CSQLite package exports encryption support as a SwiftPM package trait.
- So ideally our Swift SDK would have the same trait and then depend on CSQLite, forwarding the trait. This is not supported by SwiftPM though.
- And even if it were supported, XCode doesn't support setting package traits on dependencies. 

As a workaround, users have to:

- Add a local `Package.swift` setup somewhere in their project sources.
- In that project, depend on CSQLite with the encryption trait.
- Then depend on that local project from XCode. Since the CSQLite package can only appear once in the entire dependency graph, this makes XCode enable the trait.
- Add an initializing statement to configure encryption.

This also adds a new Demo project following those steps.

### DIFF
--- a/Demos/SwiftEncryptionDemo/SwiftEncryptionDemo/ContentView.swift
+++ b/Demos/SwiftEncryptionDemo/SwiftEncryptionDemo/ContentView.swift
@@ -1,13 +1,28 @@
 import SwiftUI
 import PowerSync
 
+enum DatabaseOpenState {
+    case loading
+    case opened(String)
+    case error(Error)
+}
+
 struct ContentView: View {
+    @State var state = DatabaseOpenState.loading
+    
     var body: some View {
         VStack {
-            Image(systemName: "globe")
-                .imageScale(.large)
-                .foregroundStyle(.tint)
-            Text("Hello, world!")
+            switch state {
+            case .loading:
+                ProgressView()
+            case .opened(let cipher):
+                Image(systemName: "lock.circle")
+                    .imageScale(.large)
+                    .foregroundStyle(.tint)
+                Text("Opened encrypted database using \(cipher)")
+            case .error(let error):
+                Text("Error opening database: \(error.localizedDescription)")
+            }
         }
         .padding()
         .task {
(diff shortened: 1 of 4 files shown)


Sources/PowerSync/PowerSyncDatabase.swift
```diff
@@ -8,16 +8,19 @@ public let DEFAULT_DB_FILENAME = "powersync.db"
...
 ///   - dbFilename: The database filename. Defaults to "powersync.db"
 ///   - logger: Optional logging interface
+///   - initialStatements: An optional list of statements to run as the database is opened.
 /// - Returns: A configured PowerSyncDatabase instance
 public func PowerSyncDatabase(
     schema: Schema,
     dbFilename: String = DEFAULT_DB_FILENAME,
-    logger: (any LoggerProtocol) = DefaultLogger()
+    logger: (any LoggerProtocol) = DefaultLogger(),
+    initialStatements: [String] = []
```
stevensJourney: Remember to add a changelog entry for this.

Demos/SwiftEncryptionDemo/SwiftEncryptionDemo/ContentView.swift
```diff
@@ -0,0 +1,39 @@
...
+            let ps = PowerSyncDatabase(
+                schema: Schema(tables: [
+                    Table(
+                        name: "users",
+                        columns: [
+                            .text("name")
+                        ]
+                    ),
+                ]),
+                initialStatements: ["pragma key = 'my encryption key'"]
```
stevensJourney: Nit: This is a simple example, so this is probably fine, but we should consider not recommending storing the key in plain text in code. We could do this with a simple comment.

One could also potentially hint at using the keychain via the `Security` library.

stevensJourney (commented):
This looks good to me. I left some comments for some minor items I think may be outstanding.

stevensJourney (approved):
Happy with the changes :)

PR #83, opened 2025-10-23
Update Kotlin SDK and core extension

simolus3:
This makes the Swift SDK compatible with changes from https://github.com/powersync-ja/powersync-kotlin/pull/277 without adding encryption support right away.

Since SQLite is no longer statically linked into the Kotlin SDK build, we now depend on our `CSQLite` fork to compile and link SQLite statically.


stevensJourney (approved):
Happy with these changes :D

PR #86, opened 2025-10-29
Sync streams

simolus3:
This adds support for sync streams by porting the API from Kotlin:

- Calling `syncStream()` on a PowerSync database instance returns a handle that can be `subscribe()`d, returning a `SyncStreamSubscription`.
- `SyncStreamSubscription` can be used to wait for that subscription to first sync.
- All kinds of sync streams, subscribed to or not, can be passed to `SyncStatusData.forStream` to get the current status for a stream.

All of this is implemented by delegating calls to the Kotlin equivalent. This also means that it's quite light on tests (reliably testing this would require porting the approach of mocking streaming HTTP responses, but the functionality is covered by Kotlin tests.

I've also added sync streams to the demo as an opt-in. Marking them as enabled via secrets would make the demo load list items on-demand as the user navigates to them.

### DIFF
--- a/Sources/PowerSync/Kotlin/db/KotlinJsonParam.swift
+++ b/Sources/PowerSync/Kotlin/db/KotlinJsonParam.swift
@@ -26,4 +26,24 @@ extension JsonValue {
             return PowerSyncKotlin.JsonParam.Map(value: anyDict)
         }
     }
+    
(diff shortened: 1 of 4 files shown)


stevensJourney (approved):
:shipit:

PR #88, opened 2025-11-04
Remove unnecessary Task's in actor

jonkan:
A plain Task { ... } (e.g. not defered) inherits the actor’s executor and removing them doesn't change the current behavior. The fact that the functions are actor-isolated is enough to ensure the code won't run on the main actor.

**Discussion**
As long as we're fine with the file system operations being isolated to the same actor (i.e. functions will be able to block each other) this should be a simple and safe solution. The downside being that it might be less performant than some concurrent solution. Is this an issue?

I'm not sure it's trivial to implement an equally safe solution that supports parallell operations. E.g. what about related operations on/in the same file or directory?

Thoughts/input?

### DIFF
--- a/Sources/PowerSync/attachments/FileManagerLocalStorage.swift
+++ b/Sources/PowerSync/attachments/FileManagerLocalStorage.swift
@@ -13,85 +13,71 @@ public actor FileManagerStorageAdapter: LocalStorageAdapter {
     }
 
     public func saveFile(filePath: String, data: Data) async throws -> Int64 {
-        return try await Task {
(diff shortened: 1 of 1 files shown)


simolus3:
Given that these operations are only called from a single asynchronous function in `SyncingService.handleSync` that doesn't setup an outer task group / concurrency between attachments, I think we didn't have any concurrency here to begin with and these changes should be fine.

I suuppose if we wanted to add concurrency for syncing attachments, `handleSync` would be a better place to add it anyway

PR #89, opened 2025-11-06
Ability to Delete SQLite files on Close

stevensJourney:
# Overview

Users have requested for the PowerSync SDK to be able to delete the SQLite files it created. 

This adds the ability to delete potentially opened files when calling `close` on the PowerSync client. (We should only delete the files once we've closed active connections).

This code could have been implemented in the Kotlin SDK and used here. I've opted not to do this for now, since updating the Kotlin SDK currently requires other changes. Implementing this functionality in Swift directly allows for a faster release cycle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,17 @@
 ## 1.6.1 (unreleased)
 
 * Update Kotlin SDK to 1.7.0.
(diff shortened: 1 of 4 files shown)


simolus3 (approved):
Happy with these changes 👍

PR #111, opened 2026-02-06
feat: Add dbDirectory parameter for custom database storage locations

nightfall708:
## Summary

- Adds a `dbDirectory` parameter to `PowerSyncDatabase()` that allows specifying a custom directory path for the database file
- Enables storing the database in a shared **App Group container**, which is essential for iOS apps that need to share data between the main app and extensions (e.g. Share Extensions, widgets)
- The underlying Kotlin SDK already supports `dbDirectory` — this change threads it through the Swift API layer

## Motivation

Currently, `dbFilename` only accepts a filename (not a path), and the database is always stored in the hardcoded Application Support directory. This makes it impossible to share the database between the main app and app extensions without going through a cloud round-trip.

**Before:** Share Extension writes to its own PowerSync database → syncs to cloud → main app syncs from cloud (slow, requires connectivity)

**After:** Both main app and Share Extension point to the same database file in the App Group container (instant local access)

## Usage

```swift
let containerURL = FileManager.default.containerURL(
    forSecurityApplicationGroupIdentifier: "group.com.example.app"
)

let db = PowerSyncDatabase(
    schema: schema,
    dbFilename: "powersync.db",
    dbDirectory: containerURL?.path
)
```

## Changes

| File | Change |
|------|--------|
| `PowerSyncDatabase.swift` | Add `dbDirectory: String? = nil` parameter to the public API |
| `KotlinPowerSyncDatabaseImpl.swift` | Thread `dbDirectory` through to the Kotlin SDK; use it in `deleteDatabase()` |
| `KotlinPowerSyncDatabaseImplTests.swift` | Add tests for custom directory creation, file location, and deletion |
| `CHANGELOG.md` | Document the new feature |

## Test plan

- [x] Verify build compiles (`swift build` ✅)
- [x] Verify tests compile (`swift build --build-tests` ✅)
- [ ] `testCustomDbDirectory` — verifies database is created in custom directory, not the default
- [ ] `testCustomDbDirectoryCloseWithDeleteDatabase` — verifies all SQLite files (db, wal, shm) are cleaned up from custom directory
- [ ] Existing tests continue to pass (default behavior unchanged when `dbDirectory` is `nil`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.11.0 (unreleased)
 
 * Add support for [sync streams](https://docs.powersync.com/sync/streams/overview).
+* Add `dbDirectory` parameter to `PowerSyncDatabase()` to support custom database storage locations. This enables storing the database in a shared App Group container for access from Share Extensions and other app extensions.
 
 ## 1.10.0
 
--- a/Sources/PowerSync/Kotlin/KotlinPowerSyncDatabaseImpl.swift
+++ b/Sources/PowerSync/Kotlin/KotlinPowerSyncDatabaseImpl.swift
@@ -11,18 +11,18 @@ final class KotlinPowerSyncDatabaseImpl: PowerSyncDatabaseProtocol,
     private let encoder = JSONEncoder()
     let currentStatus: SyncStatus
     private let dbFilename: String
+    private let dbDirectory: String?
 
     init(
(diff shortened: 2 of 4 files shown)


nightfall708:
After reconsideration. It's obvious this way doesn't make much sense. However, this means saving via extension can't really work. We need a workaround for that case.

PR #127, opened 2026-04-15
Implement CRUD transactions in Swift instead of Kotlin

simolus3:
Methods returning items from `ps_crud` (`getCrudBatch()`, `getCrudTransactions()` and `getNextCrudTransaction()`) don't do anything special: They read some items from `ps_crud` via `getAll`, marking a batch as completed removes the crud items and updates a target write checkpoint in the `$local` bucket. They don't need any integration with database internals.

So, implementing these methods in Swift is an easy first step towards eventually removing the Kotlin framework. This also allows us to provide typed values for `opData` and `previousValues`, previously they both had to be a `[String: String?]` map due to interop limitations.


Sources/PowerSync/Protocol/db/CrudBatch.swift
```diff
@@ -1,24 +1,47 @@
...
+    }
 
     /// Call to remove the changes from the local queue, once successfully uploaded.
     ///
     /// `writeCheckpoint` is optional.
-    func complete(writeCheckpoint: String?) async throws
-}
+    public func complete(writeCheckpoint: String?) async throws {
+        let lastId = crud.last!.clientId
+        try await completeCrudItems(self.db, lastId)
```
stevensJourney: It looks like this should pass in the `writeCheckpoint` param from the function call.

Sources/PowerSync/Protocol/db/CrudEntry.swift
```diff
@@ -51,13 +53,96 @@ public protocol CrudEntry: Sendable {
...
+            case .string(let value):
+                return value
+            case .int(let value):
+                return String(value)
+            case .double(let value):
+                return String(value)
+            case .bool(let value):
+                return String(value)
+            case .null:
+                return "null"
```
stevensJourney: Should this be the `"null"` string or `nil`? It might be a good idea to add a comment for this case which could be confusing in the future.

Sources/PowerSync/Protocol/db/CrudEntry.swift
```diff
@@ -51,13 +53,96 @@ public protocol CrudEntry: Sendable {
...
+            let id: String
+            let op: UpdateType
+            let data: JsonParam?
+            let type: String
+            let metadata: String?
+            let old: JsonParam?
+        }
+
+        let decoder = JSONDecoder()
+        let entry = try decoder.decode(CrudJsonEntry.self, from: data.data(using: .utf8)!)
```
stevensJourney:
```suggestion
guard let jsonData = data.data(using: .utf8) else {
    throw PowerSyncError.operationFailed(message: "Invalid UTF-8 in CRUD entry")
}

let entry = try JSONDecoder().decode(CrudJsonEntry.self, from: jsonData)
```

It's probably not a reachable path, but it would look slightly cleaner to guard against the possible `nil`

Sources/PowerSync/Protocol/db/CrudBatch.swift
```diff
@@ -1,24 +1,47 @@
 import Foundation
 
-/// A transaction of client-side changes.
-public protocol CrudBatch: Sendable {
+/// A collection of client-side changes.
+public struct CrudBatch: Sendable {
```
stevensJourney, line 4: I noticed the change from public CRUD protocols to concrete SDK-owned structs. Since these structs do not expose public initializers, users can no longer manually create `CrudEntry`, `CrudBatch`, `CrudTransaction`, or `CrudTransactions` values outside the SDK. We may not expect many users to be doing that, and it may not have been a recommended pattern, but some test code could (potentially) still be instantiating or mocking these types for upload-related logic. Given that this removes that ability, we should probably add a changelog note calling out the public API behavior change so it is visible.

stevensJourney (commented):
I'm generally happy with these changes. The typed crud entries are a very nice addition.  I left some minor comments for some items that caught my eye.

Sources/PowerSync/Protocol/db/CrudBatch.swift
simolus3: Good catch, I've also added a test for that case

Sources/PowerSync/Protocol/db/CrudEntry.swift
simolus3: It should actually be `nil` which is what the Kotlin SDK would have returned as well 👍

stevensJourney (approved):

PR #128, opened 2026-04-16
Drive sync client from Swift

simolus3:
This replaces the Kotlin+Rust sync client with a Swift+Rust sync client. We want to remove the Kotlin dependency eventually, and this is a big step towards that (the only thing left afterwards is to implement connection pools in Swift afterwards). By using the Rust client, we can implement sync over the Kotlin SDK by running `powersync_control` invocations. The Kotlin SDK is only used as a connection pool at this point.

TODOs:

- [x] Test CRUD uploads.
- [x] Restore support for Sync Streams
- [x] Restore support for raw tables.
- [x] Restore support for network loggers.
- [x] Use correct user agent.
- [x] Support old platform versions again (so we'd have to polyfill `Synchronization`).

These two items are very tricky to implement with the Kotlin database implementation. I'll tackle them in a follow-up PR replacing Kotlin entirely:

- [ ] Resolve offline sync status (done in https://github.com/powersync-ja/powersync-swift/pull/130).
- [ ] Add mutex groups to ensure two databases with the same name can't sync concurrently (done in https://github.com/powersync-ja/powersync-swift/pull/130).


Sources/PowerSync/Implementation/sync/HttpClient.swift
```diff
@@ -0,0 +1,183 @@
...
+}
+
+struct PlatformHttpClient: HttpClient {
+    let session: URLSession
+    
+    func receiveSyncLines(request: URLRequest) async throws -> (HTTPURLResponse, any SyncLineResponse) {
+        let (bytes, response) = try await session.bytes(for: request)
+        let jsonStreamMimeType = "application/x-ndjson"
+        
+        if response.mimeType != jsonStreamMimeType {
```
stevensJourney, line 49: Will the response of a `401` also be `application/x-ndjson`? If not, it looks like this might swallow `401` errors and miss invalidating credentials?

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -0,0 +1,452 @@
...
+            async let download: () = downloadLoop(signals: signals)
+            async let upload: () = uploadLoop(signals: signals)
+            
+            let _ = try await (download, upload)
+        }
+    }
+
+    private func uploadLoop(signals: SyncSignals) async throws {
+        // TODO: Replace with better watch mechanism once we've dropped the Kotlin dependency and can use onChange.
+        let watch = try db.watch(sql: "SELECT 1 FROM ps_crud LIMIT 1", parameters: [], mapper: { _ in () })
```
stevensJourney: We might want to throttle the watch query here by `crudThrottle` provided in the options.

stevensJourney (commented):
I only had some minor comments/questions. Overall I'm happy with the approach taken here.

Sources/PowerSync/Implementation/sync/HttpClient.swift
simolus3: Good catch! I've added a `UnexpectedResponseError` for this case which allows the sync client to invalidate credentials.

stevensJourney (approved):

PR #129, opened 2026-04-20
Implement schema serialization in Swift

simolus3:
The core extension expects a JSON-serialized schema to setup tables, views, triggers and the sync client. Currently, the Swift SDK maps its table structures to Kotlin which is then responsible for the serialization.

This implements the necessary steps in Swift through the `Encodable` protocol. This is not that useful on its own because we still have to go through Kotlin, but this will be helpful as we eventually remove the Kotlin dependency here.

### DIFF
--- a/Sources/PowerSync/Protocol/Schema/RawTable.swift
+++ b/Sources/PowerSync/Protocol/Schema/RawTable.swift
@@ -11,7 +11,7 @@
 ///
 /// Note that raw tables are only supported when ``ConnectOptions/newClientImplementation``
 /// is enabled.
-public struct RawTable: BaseTableProtocol {
+public struct RawTable: BaseTableProtocol, Encodable {
     /// The name of the table as it appears in sync rules.
     ///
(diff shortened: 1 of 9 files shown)


Sources/PowerSync/Protocol/Schema/RawTable.swift
```diff
@@ -58,11 +60,52 @@ public struct RawTable: BaseTableProtocol {
...
+        let encoder = JSONEncoder()
+        do {
+            let serialized = try encoder.encode(self)
+            return String(data: serialized, encoding: .utf8)!
+        } catch {
+            // An older version of the Swift SDK used to implement this method in Kotlin.
+            // That could also throw an exception (which would crash the process), but that
+            // was overlooked due to a missing @Throws annotation.
+            // Now, we can't mark this as throws without breaking backwards compatibility.
+            // We should conver this to be throwing in a future major release.
```
stevensJourney:
```suggestion
            // We should convert this to be throwing in a future major release.
```

stevensJourney (dismissed):
Happy with these changes :)

stevensJourney (approved):

PR #132, opened 2026-05-15
Enable zizmor

simolus3:
This enables zizmor as a static analysis tool for GitHub actions, using the same configuration as the [Dart SDK](https://github.com/powersync-ja/powersync.dart/pull/416).

This mainly involves setting `permissions: {}` for most jobs and adding a concurrency limit. Also, the release process has been changed a bit:

1. We don't need to push a tag manually since `gh release create` is supposed to do that ([docs](https://cli.github.com/manual/gh_release_create), "If a matching git tag does not yet exist, one will automatically get created from the latest state of the default branch.").
2. Zizmor wants us to use `gh release` directly instead of `ncipollo/release-action`.

While refactoring the release process, I realized we have the option of doing pre-releases! So, this also prepares a beta release for 1.14.0, which we can release without waiting for potential GRDB changes we might want to add as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.14.0 (unreleased)
+## 1.14.0-Beta.0
 
 * Remove internal dependency on the PowerSync Kotlin SDK. Going forward, the Swift SDK is implemented in Swift!
   __Important__: While these changes are tested, they are a full rewrite of the internal connection pool logic.
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.14.0 (unreleased)
+## 1.14.0-Beta.0
 
 * Remove internal dependency on the PowerSync Kotlin SDK. Going forward, the Swift SDK is implemented in Swift!
```
stevensJourney, line 5: It might be nice to add a small note as to why this is being released as a pre-release. But, maybe it's not necessary.

stevensJourney (approved):

CHANGELOG.md
simolus3: I'll leave it as-is for now, we'll probably skip the pre-release anyway and bump this to 1.14.0 when we ship this.

PR #133, opened 2026-05-15
Test Sync Stream with object/array parameters

simolus3:
The issue reported in https://github.com/powersync-ja/powersync-kotlin/issues/349 doesn't affect the Swift implementation of the SDK anymore, but it's still good to add a test. So, this copies the test from https://github.com/powersync-ja/powersync-kotlin/pull/350.


stevensJourney (approved):
Wonderful!

PR #137, opened 2026-05-26
Await initialization before connecting

### LINKED ISSUE #135: [BUG] (?) `PowerSyncDatabaseImpl.connect` should initialize powersync
I believe that this function needs to be calling `initialize()` before attempting the connection - similar to the other methods in this class.

https://github.com/powersync-ja/powersync-swift/blob/6e5eaeb6b15f1ddba05269816220faf7ad7027cb/Sources/PowerSync/Implementation/PowerSyncDatabaseImpl.swift#L129-L131

I'm running into issues where my attempt to connect is racing with the initialization process such that when the StreamingSyncClient attempts to upload the crud queue, the `ps_crud` table still hasn't been created yet. Because of this the `try db.watch(options: <select from ps_crud>)` in the `uploadLoop` method throws an error and causes the uploadLoop to fail - it doesn't get restarted. So for the remainder of the application lifetime, no uploading occurs.

https://github.com/powersync-ja/powersync-swift/blob/6e5eaeb6b15f1ddba05269816220faf7ad7027cb/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift#L29-L45

I believe the pathological reason for this is that I am leveraging the GRDB integration with the PowerSyncDatabase so my applications writes to the database don't go through the methods on PowerSyncDatabaseImpl that call `initialize`.

I'm not sure if my usage of the `connect()` method to startup the PowerSync client service is the proper way to wield the client - so I'm open to suggestions if I'm just holding it wrong.

simolus3:
As reported in https://github.com/powersync-ja/powersync-swift/issues/135, the current implementation has a race condition where we might connect before the database is fully initialized, causing errors due to missing internal tables.

This fixes that in two ways:

1. In `connect()`, explicitly call `try await initialize()` first.
2. Change the implementation of `watch()` to be based on `PowerSyncDatabaseImpl` methods instead of the underlying connection pool. This ensures that e.g. `getQuerySourceTables` calls the `getAll` implementation on the database, which will initialize it too.

This also removes the beta suffix to prepare a release.

Closes https://github.com/powersync-ja/powersync-swift/issues/135.


Sources/PowerSync/Implementation/queries/watch.swift
```diff
@@ -0,0 +1,133 @@
+import AsyncAlgorithms
+
+func watchImpl<RowType: Sendable>(db: PowerSyncDatabaseImpl, options: WatchOptions<RowType>) -> AsyncThrowingStream<[RowType], any Error> {
+    // Note: Hold a weak reference to the database, active watch streams shouldn't prevent a database from being closed.
+    AsyncThrowingStream { continuation in
+        // Create an outer task to monitor cancellation
+        let task = Task { [weak db] in
```
simolus3, line 7: This is the same implementation we had before, except that it uses a weak reference to `PowerSyncDatbaaseImpl` instead of a strong reference to the underlying pool. I've also refactored `continuation.finish()` to be called in a `defer` block, ensuring wee call it whenever the task returns.

Sources/PowerSync/Implementation/queries/watch.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        // Propagate cancellation from the outer task to the inner task
+        continuation.onTermination = { @Sendable _ in
+            task.cancel() // This cancels the inner task when the stream is terminated
+        }
+    }
+}
+
+private func prepareWatch(
```
stevensJourney: Nit: looks like this is unused and can be removed.

stevensJourney (dismissed):
Happy with these changes

Sources/PowerSync/Implementation/queries/watch.swift
simolus3: Good catch, thanks!

stevensJourney (approved):

PR #144, opened 2026-06-08
Fix `readTransaction` throwing for connection pools

### LINKED ISSUE #142: SQLite error “attempt to write a readonly database” triggered by readTransaction (BEGIN IMMEDIATE) in PowerSync Swift SDK
## 🐛 Issue: readTransaction incorrectly uses BEGIN IMMEDIATE

**Summary**

When using `readTransaction` in the PowerSync Swift SDK, the internal transaction implementation executes:

```sql
BEGIN IMMEDIATE
```

This causes a runtime failure in some cases with:

```text
sqliteError(extendedResultCode: 8, message: "attempt to write a readonly database", sql: "BEGIN IMMEDIATE")
```

Even though the API is explicitly intended for read-only access, the transaction runner always starts a write-capable transaction mode (`IMMEDIATE`), which can fail depending on the SQLite connection state, lock conditions, or read-only restrictions of the underlying database context.

---

**Steps to reproduce**

```swift
try await db.readTransaction { tx in
    return try tx.get(
        sql: "SELECT * FROM products WHERE product_cic = ?",
        parameters: [productCic]
    ) { cursor in
        try Product(
            id: UUID(uuidString: cursor.getString(name: "id"))!,
            name: cursor.getString(name: "product_name")
        )
    }
}
```

---

**Expected behavior**

`readTransaction` should not require write-level locking and should not execute:

```sql
BEGIN IMMEDIATE
```

It should use a read-safe transaction mode such as:

- `BEGIN`
- `BEGIN DEFERRED`
- or no explicit transaction when not required

---

**Actual behavior**

The SDK executes:

```sql
BEGIN IMMEDIATE
```

inside:

```swift
TransactionImpl.run(conn:callback:)
```

which leads to:

```text
sqliteError(extendedResultCode: 8, message: "attempt to write a readonly database")
```

---

**Impact**

- Affects all `readTransaction`, `get`, and `getAll` queries
- Can break read flows in SwiftUI `.task` / `.onAppear`
- Occurs even when no write operations are performed

---

**Notes**

The issue appears to come from using the same transaction runner (`TransactionImpl.run`) for both read and write operations, always defaulting to:

```sql
BEGIN IMMEDIATE
```

regardless of transaction type.

---

**Environment**

- PowerSync Swift SDK: 1.14.1
- iOS app using SwiftUI
- SQLite-backed local database initialized via PowerSync

simolus3:
Matching the behavior from our Dart and JavaScript SDKs, the Swift SDK started using `BEGIN IMMEDIATE` for read transactions in version 1.14.0. The reason for this is that we want to start the transaction as soon as possible, instead of having the WAL cutoff at the first statement used in the callback.

Unfortunately, the Swift SDK also enables `pragma query_only = TRUE` on read connections, which prevents `BEGIN IMMEDIATE`. This is not actually necessary though, we also open connections with `SQLITE_READONLY` to prevent writes. So, the fix is to simply not do that.

This also adds a test verifying that read transaction work on connection pools (the default tests with in-memory database don't catch this issue) and that we can't write in them.

Closes #142.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.14.2
+
+* Fix `readTransaction` throwing. This issue was introduced in version 1.14.0.
+
 ## 1.14.1
 
 * [Internal] The GRDB integration uses high-level GRDB APIs to run statements.
--- a/Sources/PowerSync/CurrentVersion.swift
+++ b/Sources/PowerSync/CurrentVersion.swift
@@ -1,2 +1,2 @@
 // The current version of the PowerSync Swift SDK. This should be updated to the latest version in `CHANGELOG.md` when a new version is released.
-let libraryVersion = "1.14.1"
+let libraryVersion = "1.14.2"
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -90,9 +90,6 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
 
         if isWriter {
             let _ = try context.execute(sql: "pragma journal_mode = WAL", parameters: [])
-        } else {
-            // This is mainly an additional safety element, we also open read connections SQLITE_READONLY.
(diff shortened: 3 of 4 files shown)


Chriztiaan (approved):
Looks good to me.

PR #147, opened 2026-06-15
Multi-process database support: absolute paths, open retry, cross-process change signal

asiergmorato:
## What

Three core changes that let a PowerSync database file be safely shared across processes (an app and its widgets or App Intents extensions). Extracted from #146 (the SwiftData `DataStore`) for separate review, as requested there.

This PR has no SwiftData dependency: it only touches the `PowerSync` target. #146 will rebase on top of it.

## Changes

1. **Absolute `dbFilename` paths.** `PowerSyncDatabase(dbFilename:)` treats a path starting with `/` as absolute and uses it as-is, so the database can live in an App Group container. `close(deleteDatabase: true)` deletes the files at that location. Plain filenames keep the existing behavior (default databases directory).

2. **Concurrent-open retry.** `pragma journal_mode = WAL` needs a moment of exclusive access and reports `SQLITE_BUSY`/`SQLITE_BUSY_RECOVERY` *without consulting the busy handler*, so `busy_timeout` cannot cover the open path. Two processes opening the same file concurrently (an app launching while its extension opens the same file) failed ~80% of the time in a two-process harness, even for read-only consumers (the pool always opens a writer first). Opening and configuring a connection now retries with backoff. Reproduced in-process with a system-SQLite connection holding a reserved lock, which is the same file-level contention two processes produce.

3. **Cross-process change signal.** Update hooks are per-pool, so `watch` was structurally blind to writes made by other processes. Each pool now posts a Darwin notification after every committed write (the mechanism Core Data uses for its remote-change notifications; name derived from the canonical file path) and, on receipt, re-emits `tableUpdates` with `EXTERNAL_CHANGES_MARKER`, which `watch` and the upload client treat as "unknown tables changed". In-memory databases skip the signal. Two `PowerSyncDatabase` instances over the same file in one process use independent pools, reproducing the cross-process behavior deterministically (this is how the test exercises it).

## On locking

You mentioned an asynchronous locking scheme in your other SDKs to avoid using the write connection concurrently. The open retry here is a pragmatic first step (it backs off on `SQLITE_BUSY` during the WAL transition, where the busy handler does not apply), and it currently sleeps on a GCD thread rather than awaiting. I'm happy to align this with your existing scheme instead of keeping a parallel mechanism: where does it live? Same question for guarding concurrent *writes* across processes (today they serialize through `busy_timeout`).

## Testing

Builds with `-strict-concurrency=complete`. 35 tests pass across the new suites (`ConcurrentOpenTests`, `CrossProcessSignalTests`, `AbsolutePathTests`) and the in-memory sync integration tests. Also validated with a real two-process harness (separate OS processes over one file): concurrent cold opens 5/5 clean, and the observing process sees writes made by the other process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## Unreleased
+
+* `PowerSyncDatabase(dbFilename:)` now accepts an absolute path (starting with `/`), used
+  as-is so the database can live in an App Group container shared with app extensions.
+  Plain filenames keep the existing behavior, and `close(deleteDatabase: true)` deletes the
+  files at the absolute location.
+* Opening the connection pool retries while another process holds the database. The
+  `pragma journal_mode = WAL` transition reports `SQLITE_BUSY`/`SQLITE_BUSY_RECOVERY`
+  without consulting the busy handler, so concurrent cold opens (an app launching while its
+  extension opens the same file) used to fail; the pool now retries with backoff.
+* Added an opt-in cross-process change signal: each pool posts a Darwin notification after
+  every committed write and, on receipt, re-emits `tableUpdates` with
+  `EXTERNAL_CHANGES_MARKER` so `watch` queries and the upload client wake for writes made by
+  other processes sharing the database file. In-memory databases skip the signal.
+
 ## 1.14.3
 
 - Fix CRUD uploads entering a `Delaying due to previously encountered CRUD item` loop.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@
 * Added an opt-in cross-process change signal: each pool posts a Darwin notification after
   every committed write and, on receipt, re-emits `tableUpdates` with
   `EXTERNAL_CHANGES_MARKER` so `watch` queries and the upload client wake for writes made by
-  other processes sharing the database file. In-memory databases skip the signal.
+  other processes sharing the database file. Only databases opened from an absolute path
+  (an App Group container) use the signal; in-memory and default-directory databases skip
+  it, since they can't be shared across processes.
 
 ## 1.14.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,8 @@
 
 * `PowerSyncDatabase(dbFilename:)` now accepts an absolute path (starting with `/`), used
   as-is so the database can live in an App Group container shared with app extensions.
-  Plain filenames keep the existing behavior, and `close(deleteDatabase: true)` deletes the
-  files at the absolute location.
-* Opening the connection pool retries while another process holds the database. The
-  `pragma journal_mode = WAL` transition reports `SQLITE_BUSY`/`SQLITE_BUSY_RECOVERY`
-  without consulting the busy handler, so concurrent cold opens (an app launching while its
-  extension opens the same file) used to fail; the pool now retries with backoff.
-* Added an opt-in cross-process change signal: each pool posts a Darwin notification after
-  every committed write and, on receipt, re-emits `tableUpdates` with
-  `EXTERNAL_CHANGES_MARKER` so `watch` queries and the upload client wake for writes made by
-  other processes sharing the database file. Only databases opened from an absolute path
-  (an App Group container) use the signal; in-memory and default-directory databases skip
-  it, since they can't be shared across processes.
+  Plain filenames keep the existing behavior. The SDK coordinates opening the database to
+  avoid conflicts and can share update notifications across the main app and extensions.
 
 ## 1.14.3
 
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -5,7 +5,20 @@ import Foundation
 enum DatabaseLocation {
     case inMemory
     case inDefaultDirectory(name: String)
-    
+    case atPath(String)
+
+    /// The on-disk path other processes can share, or `nil` for in-memory databases.
+    var sharedPath: String? {
+        switch self {
+        case .inMemory:
+            return nil
+        case let .inDefaultDirectory(name):
+            return (try? DatabaseLocation.appleDefaultDatabaseDirectory().path).map { "\($0)/\(name)" }
+        case let .atPath(path):
+            return path
+        }
+    }
+
     func openConnection(writer: Bool) throws -> RawSqliteConnection {
         var db: OpaquePointer?
         let rc: Int32
@@ -24,6 +37,21 @@ enum DatabaseLocation {
             }
 
             path = "\(databaseDirectory)/\(name)"
+            let flags = if writer {
+                SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE
+            } else {
+                SQLITE_OPEN_READONLY
+            }
+            rc = sqlite3_open_v2(path, &db, flags, nil)
+        case .atPath(let absolutePath):
+            let fileManager = FileManager.default
+            let directory = (absolutePath as NSString).deletingLastPathComponent
+
+            if !fileManager.fileExists(atPath: directory) {
+                try fileManager.createDirectory(atPath: directory, withIntermediateDirectories: true)
+            }
+
+            path = absolutePath
             let flags = if writer {
                 SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE
             } else {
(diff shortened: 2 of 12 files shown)


Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -30,6 +43,21 @@ enum DatabaseLocation {
...
+                try fileManager.createDirectory(atPath: directory, withIntermediateDirectories: true)
+            }
+
+            path = absolutePath
+            let flags = if writer {
+                SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE
+            } else {
+                SQLITE_OPEN_READONLY
+            }
+            rc = sqlite3_open_v2(path, &db, flags, nil)
```
simolus3: Can we introduce a helper function to share this with the `inDefaultDirectory` case?

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -110,6 +146,42 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+        guard case let PowerSyncError.sqliteError(extendedResultCode, _, _, _, _) = error else {
+            return false
+        }
+        let primary = extendedResultCode & 0xFF
+        return primary == SQLITE_BUSY || primary == SQLITE_LOCKED
+    }
+
+    /// Opens and configures a connection, retrying with backoff while another process
+    /// holds the database (apps and their widgets/extensions open concurrently).
+    fileprivate func openConfiguredConnection(writer: Bool) throws -> RawSqliteConnection {
```
simolus3: This isn't that bad, so only do it if it doesn't require too many changes: Can we make this function asynchronous and use `Task.sleep`  instead of blocking a thread? That may also help for cases where a task using a database currently being opened gets cancelled.

I imagine we can do one `runBlocking` call to open the connection without running any pragma statements (that shouldn't really do IO anyway), then call `configureConnection` in a loop like this, using one `runBlocking` call per attempt and delaying asynchronously.

Sources/PowerSync/Protocol/SQLiteConnectionPool.swift
```diff
@@ -57,6 +57,11 @@ public protocol SQLiteStatementIteratorProtocol {
 /// This is the underlying pool implementation on which the higher-level PowerSync Swift SDK is built on.
 /// 
 /// This is an internal protocol and should not be implemented outside of the PowerSync SDK.
+/// Emitted on `tableUpdates` when another process changed the database. The concrete
+/// tables are unknown (cross-process signals carry no payload), so consumers must treat
+/// this as potentially matching every table they watch.
+public let EXTERNAL_CHANGES_MARKER = "__powersync_external_changes__"
```
simolus3: I think this should be internal. We don't have an `onChange` API here, and `watch` already takes this into account. So unless I'm missing anything, I don't see why users would care about this?

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -62,11 +90,16 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     private let logger: any LoggerProtocol
     private let tableUpdatesStream = BroadcastStream<Set<String>>()
     private let opener = PoolOpener()
+    /// Cross-process change signaling; `nil` for in-memory databases (nothing to share).
+    private let changeSignal: CrossProcessChangeSignal?
```
simolus3, line 79: Realistically, `inDefaultDirectory` is also not shareable with app extensions, right? I wonder if we should only enable this for `.atPath` then.

simolus3 (commented):
I did some research on what Darwin APIs there are for IPC, and it looks like the good stuff is all inaccessible to apps. So I think synchronizing by retrying on `sqlite_busy` errors is actually fine. It might be worth exerimenting with [this suggestion](https://developer.apple.com/documentation/foundation/improving-performance-and-stability-when-accessing-the-file-system#Coordinate-file-access-to-reduce-hangs) to manage exclusive file access through `NSFileCoordinator`, but that sounds like it could add a fair bit of overhead.

The recovery loop when we open databases looks good to me (modulo blocking a thread, but even that it likely fine). For write statements, relying on `sqlite_busy` initially sounds acceptable to me too.

asiergmorato:
Thanks  @simolus3, all addressed in the latest push.

- **Async retry**: moved the backoff loop off the blocking usleep to Task.sleep, so it no longer pins a thread and a database open is now cancellable (added a test for that). I kept the whole pool building inside one runBlocking call rather than splitting open-then-configure, because RawSqliteConnection is ~Copyable and can't cross the async boundary; the retry/cancellation benefit comes from awaiting between attempts, which this still gets.
- **Signal only for .atPath**: good catch, the default directory is in the app sandbox and not reachable by extensions, so I now only enable the change signal for absolute-path databases.
- **EXTERNAL_CHANGES_MARKER**: made it internal.
- **Shared open helper and the isDirectory**: hint: done.

On the article: I read through it. Most of it targets document-style access (iCloud Drive, file providers), and NSFileCoordinator specifically doesn't fit SQLite in WAL mode well: SQLite already coordinates concurrency with POSIX file locks, WAL spans three files (.sqlite, -wal, -shm) plus shared-memory mmap that a single-URL coordinator doesn't understand, and Core Data deliberately doesn't use file coordination for its SQLite store. So I think retrying on SQLITE_BUSY (as you concluded) is the right mechanism here. I did adopt the parts that apply to us: the isDirectory: hint to avoid a blocking stat, and all our database I/O already runs off the main thread on background GCD queues, which the async retry reinforces.

Sources/PowerSync/Implementation/CrossProcessChangeSignal.swift
```diff
@@ -0,0 +1,75 @@
...
+final class CrossProcessChangeSignal: @unchecked Sendable {
+    private let name: String
+    private let logger: any LoggerProtocol
+    private var token: Int32 = NOTIFY_TOKEN_INVALID
+    private let queue = DispatchQueue(label: "powersync.cross-process-signal")
+
+    init(databasePath: String, logger: any LoggerProtocol) {
+        // Stable across processes: both sides derive the name from the canonical path.
+        let canonical = URL(fileURLWithPath: databasePath).standardizedFileURL.path
+        self.name = "co.powersync.changes.\(Self.fnv1a(canonical))"
```
simolus3:
```suggestion
        self.name = "com.powersync.changes.\(Self.fnv1a(canonical))"
```

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+* Opening the connection pool retries while another process holds the database. The
+  `pragma journal_mode = WAL` transition reports `SQLITE_BUSY`/`SQLITE_BUSY_RECOVERY`
+  without consulting the busy handler, so concurrent cold opens (an app launching while its
+  extension opens the same file) used to fail; the pool now retries with backoff.
+* Added an opt-in cross-process change signal: each pool posts a Darwin notification after
+  every committed write and, on receipt, re-emits `tableUpdates` with
+  `EXTERNAL_CHANGES_MARKER` so `watch` queries and the upload client wake for writes made by
+  other processes sharing the database file. Only databases opened from an absolute path
+  (an App Group container) use the signal; in-memory and default-directory databases skip
+  it, since they can't be shared across processes.
```
simolus3: We don't need to point out that two features are compatible, that should be the default expectation :D

Also, the last two points are internals which we don't need to describe in changelogs in depth.

```suggestion
* `PowerSyncDatabase(dbFilename:)` now accepts an absolute path (starting with `/`), used
  as-is so the database can live in an App Group container shared with app extensions.
  Plain filenames keep the existing behavior. The SDK coordinates opening the database to avoid conflicts and can share update notifications across the main app and extensions.
```

simolus3 (dismissed):
This looks good to me apart from some minor comments now 👍 Thanks for your contribution!

Sources/PowerSync/PowerSyncDatabase.swift
```diff
@@ -6,7 +6,10 @@ public let DEFAULT_DB_FILENAME = "powersync.db"
 /// Creates a PowerSyncDatabase instance
 /// - Parameters:
 ///   - schema: The database schema
-///   - dbFilename: The database filename. Defaults to "powersync.db"
+///   - dbFilename: The database filename. Defaults to "powersync.db". Plain names are
+///     stored in the default databases directory; an absolute path (starting with "/") is
+///     used as-is, which allows sharing the database with app extensions through an App
```
simolus3, line 11: Also expand this comment to mention that while the database itself can be used concurrently across app extensions, an extension must not call `connect`.

Two connections on the same database can cause trouble, it's at least a waste of resources and is also not tested (I wouldn't be surprised if there's a way to corrupt the sync client that way).

asiergmorato:
Thanks! All three applied in the latest push:

- Renamed the Darwin notification prefix to com.powersync.changes.
- Simplified the changelog entry to your suggested wording.
- Expanded the dbFilename doc to note that the database can be used concurrently from the app and its extensions, but only the main app should call connect (extensions read and write, syncing stays with the app).

simolus3 (approved):

simolus3:
Thank you for your contribution! Before we release this, we'll need to update our documentation website to point out potential issues (like `connect()` across app extensions not being allowed, and problems that can arise when apps using different versions of the SDK share a database). We'll take care of that, and we'll also reach out about the Swift Data integration. That will probably take a few more days, thanks for your patience.

asiergmorato:
Thank you for letting me contribute! 🤗

asiergmorato:
Quick question @simolus3, do you have any ETA on when it's going to be released the next tagged version of the main package with this?

simolus3:
Sorry for the slow response, I think we're probably looking at a release next week. A few of us are at Local-First Conf or on vacation, and it's a pretty big release.

simolus3:
We've just released version 1.15.0 of the PowerSync Swift SDK, which contains your PR.

PR #153, opened 2026-06-29
feat: Checkpoint Requests

stevensJourney:
# Overview

This is the Swift SDK implementation of the Request Checkpoint proposal: https://github.com/orgs/powersync-ja/discussions/324

Related to:

- https://github.com/powersync-ja/powersync-service/pull/696
- https://github.com/powersync-ja/powersync-sqlite-core/pull/198

This adds a new `requestCheckpoint` method which returns a `CheckpointRequest`. The `CheckpointRequest` instance exposes a `waitForSync` method which resolves once the requested checkpoint has been synced and applied locally.

```swift
func refreshFromRemote(timeout: TimeInterval = 30) async throws {
    let checkpoint = try await powerSync.requestCheckpoint()
    try await checkpoint.waitForSync(timeout: timeout)
}
```

The SDK is updated to cater for the Rust core and PowerSync service changes. `ConnectOptions` now has a `checkpointMode`, and checkpoint requests are enabled by connecting with `checkpointMode: .requests()`. The retry delay for unapplied requests is configurable via `.requests(checkpointRequestRetryDelay:)`. In request mode the sync client posts checkpoint requests to the service, tracks the latest applied checkpoint request ID from core sync status, retries unapplied requests with a delay, and cancels pending waits when the sync client disconnects or closes.

For backends that process uploads asynchronously, this also adds `CustomCheckpointRequestConnector`. A connector can implement this to handle checkpoint requests itself and return the effective checkpoint request ID accepted by the backend.

Errors thrown while creating a request (`CheckpointRequestError`) and while waiting for one (`CheckpointWaitError`) both conform to a shared `CheckpointError` protocol, so a caller driving both steps in a single `do` block can catch them together as `catch let error as any CheckpointError` while still switching on the concrete type when needed.

These APIs are documented as alpha and may change in future releases.

# Demo

The `PowerSyncExample` has been updated to include a pull-to-refresh action which shows a loading spinner until a checkpoint request has been resolved.

In order to showcase the functionality a bit better, a new Admin page has been added to the demo which conveys the SyncStatus. In the demo below, we go offline, create 50k todos which will be synced, then connect and await a checkpoint request. We can see the download progress reach 100%, but the data is only available after the checkpoint has been applied locally. The checkpoint request correctly keeps the operation in a loading state until the data is actually available.

https://github.com/user-attachments/assets/2d50175e-a513-4c96-b4bc-b79c728f03f9

There is also a small SwiftPM command-line demo under `Demos/CustomCheckpointDemo` which shows a custom checkpoint request connector against the Node.js todo backend.

# Testing

I added coverage for creating and waiting on checkpoint requests, timeout/error behavior, unsupported service responses, reconnects, seeded/reseeded checkpoint request state, retry timing, disconnect cancellation, custom checkpoint request connectors, upload integration, and the mock HTTP client hook used by these tests.

---

AI disclosure: This work was assisted by Codex and Claude Code. I manually guided and reviewed the changes.


Sources/PowerSync/Implementation/CheckpointRequestImpl.swift
```diff
@@ -0,0 +1,44 @@
...
+    init(requestId: Int64, db: any PowerSyncDatabaseProtocol) {
+        self.requestId = requestId
+        self.status = db.currentStatus
+    }
+
+    var isSynced: Bool {
+        Self.isSynced(status: status, requestId: requestId)
+    }
+
+    func waitForSync() async throws {
```
simolus3, line 30: Can we use typed throws to make matching errors easier?


```suggestion
    func waitForSync() async throws(CheckpointWaitError) {
```

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -266,14 +548,51 @@ private struct ActiveSyncIteration: Sendable {
...
+                do {
+                    let serviceEvents = try await syncClient.fetchSyncLines(request: request)
+                    // Merge the real stream, the checkpoint-validation sentinel stream and
+                    // local events into a single control loop. The validation stream never
+                    // yields control arguments: it only finishes when seeding succeeds, or
+                    // throws when seeding fails. Since AsyncAlgorithms.merge rethrows failures
+                    // from any input sequence, a seed failure still tears down this sync
+                    // iteration even while sync-line events are allowed to flow before
+                    // checkpoint request state is ready.
+                    controlArgs = AsyncAlgorithms.merge(
```
simolus3, line 557: We can use `AsyncMerge3Sequence` to avoid the nested merge call.

Demos/CustomCheckpointDemo/Sources/CustomCheckpointDemo/main.swift
```diff
@@ -0,0 +1,123 @@
+import Foundation
```
simolus3, line 1: Since it has no external dependencies, could we make this demo a (non-`product`) target in the main `Package.swift`? That would allow running it with `swift run` from the checkout directory, and makes it easier to open its sources and the main SDK at the same time in XCode.

Demos/CustomCheckpointDemo/Sources/CustomCheckpointDemo/main.swift
```diff
@@ -0,0 +1,123 @@
...
+try await db.close()
+
+private func monitorSyncErrors(_ status: any SyncStatus) -> Task<Void, Never> {
+    Task {
+        var previousDownloadError: String?
+        var previousUploadError: String?
+
+        for await update in status.asFlow() {
+            guard !Task.isCancelled else {
+                return
```
simolus3: I hope this check is not necessary, `asFlow()` should automatically clean up when cancelled.

Demos/GRDBDemo/GRDBDemo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,5 +1,5 @@
 {
-  "originHash" : "3de68e554d2ad75cf2c4dbdaf311828c8260cc6008b42e8738726e62912934ce",
+  "originHash" : "84d25347b5249e7ab78894935a203b13d9d55e5a01b7fe00745bdf028062df42",
```
simolus3, line 2: It looks like dependencies in the two existing demos would have to be re-resolved now that we've published the core extension update.

Sources/PowerSync/Implementation/sync/Instruction.swift
```diff
@@ -8,11 +8,12 @@ enum CoreLogSeverity: String, Decodable {
 enum Instruction {
     case logLine(severity: CoreLogSeverity, line: String)
     case updateSyncStatus(status: CoreDownloadSyncStatus)
-    case establishSyncStream(request: JsonParam)
+    case establishSyncStream(request: JsonParam, checkpointRequest: CheckpointRequestPayload?)
     case fetchCredentials(didExpire: Bool)
     case closeSyncStream(hideDisconnect: Bool)
     case flushFileSystem
```
simolus3: nit: This can be removed now

Sources/PowerSync/Implementation/sync/Instruction.swift
```diff
@@ -85,7 +92,15 @@ extension Instruction: Decodable {
         case .flushFileSystem:
             self = Instruction.flushFileSystem
         case .didCompleteSync:
-            self = Instruction.didCompleteSync
+            let nestedContainer = try container.nestedContainer(keyedBy: Instruction.DidCompleteSyncCodingKeys.self, forKey: .didCompleteSync)
+            self = Instruction.didCompleteSync(
+                appliedCheckpointRequestId: try nestedContainer.decodeIfPresent(
```
simolus3: This is unused now since we extract the value from the sync status, should we skip decoding this?

We could also remove it from the core extension again if we don't need this in any other SDK.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -144,7 +204,205 @@ The next upload iteration will be delayed.
...
+
+    /// Posts a checkpoint request payload prepared by core and returns the state accepted remotely.
+    private func postCheckpointRequest(_ payload: CheckpointRequestPayload) async throws -> Int64 {
+        if let customCheckpointRequestConnector {
+            do {
+                guard let requestId = Int64(payload.checkpoint_request_id) else {
+                    throw CheckpointRequestError.operationFailed(
+                        message: "Invalid checkpoint request ID received from core."
+                    )
+                }
```
simolus3: Instead of checking this here, can we make it an `Int64` on the type and decode with the string-to-int64 helper? Most use-sites explicitly have to convert their ints to a string only for this to do the reverse.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -144,7 +204,205 @@ The next upload iteration will be delayed.
...
+                underlyingError: error
+            )
+        }
+    }
+
+    /// Ensures the core checkpoint request counter has been seeded for the current stream.
+    fileprivate func seedCheckpointRequestState(checkpointRequest: CheckpointRequestPayload?) async throws {
+        guard case .requests = checkpointMode else {
+            // legacy mode does not require tracking local state
+            signals.markCheckpointsReady()
```
simolus3: Should this call `failPendingCheckpointRequests` since all checkpoint requests would fail anyway?

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -144,7 +204,205 @@ The next upload iteration will be delayed.
...
+            return
+        }
+
+        // In requests mode core is expected to always provide the payload alongside the
+        // establishSyncStream instruction. Guard instead of force-unwrapping so a mismatched
+        // core version (or protocol change) surfaces a recoverable error rather than crashing
+        // the sync task. The error routes through the normal sync-line retry path.
```
simolus3: I think crashing is the right thing to do, the core extension not providing that payload in requests mode is a programming error after all.

Retrying doesn't help because the same thing would happen on the next iteration, so failing as loud as possible sounds correct to me.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -144,7 +204,205 @@ The next upload iteration will be delayed.
...
+            try ctx.powersyncCurrentCheckpointRequestId()
+        }
+    }
+
+    private func checkpointRequestRetryLoop(signals: SyncSignals) async {
+        guard case .requests = checkpointMode else {
+            return
+        }
+
+        while !Task.isCancelled {
```
simolus3: Nit: Tasks are supposed to throw at suspension points when cancelled, so this can also be a `while true`.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -237,9 +494,27 @@ The next upload iteration will be delayed.
 
         return ControlInvocationsFromStream(sequence: stream)
     }
-    
+
     static let jsonEncoder = JSONEncoder()
     static let jsonDecoder = JSONDecoder()
+
+    private static func decodeWriteCheckpointId(from data: Data) throws -> Int64 {
+        let checkpoint = try jsonDecoder.decode(WriteCheckpointResponse.self, from: data).data.write_checkpoint
```
simolus3: These two should use `StringEncodedInt64`?

Sources/PowerSync/Implementation/sync/SyncSignals.swift
```diff
@@ -0,0 +1,207 @@
...
+final class SyncSignals: Sendable {
+    /// A pending `requestCheckpoint()` caller blocked before creating a checkpoint request ID.
+    private struct PendingCheckpointRequestWaiter {
+        let id: Int64
+        let continuation: AsyncThrowingStream<Void, any Error>.Continuation
+    }
+
+    /// Tracks connect-time validation for pending checkpoint request creation.
+    ///
+    /// Request ID allocation waits on this state so `next_checkpoint_request_id` only runs after
```
simolus3: It's not super obvious what `next_checkpoint_request_id` does here, maybe say that we have to reconcile the local counter before we're able to request checkpoints?

Sources/PowerSync/Implementation/sync/SyncSignals.swift
```diff
@@ -0,0 +1,207 @@
...
+
+        let initialWakeGeneration = pendingCheckpointRequests.withLock { $0.downloadRetryWakeGeneration }
+
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            group.addTask {
+                try await sleepForSeconds(seconds: seconds)
+            }
+
+            group.addTask {
+                let stream = self.signalPendingCheckpointRequestWaitingForReady.subscribe(bufferingPolicy: .bufferingNewest(1))
```
simolus3: Instead of this generation system, couldn't we call `subscribe` first thing in this method and then await the async iterator here? The act of subscribing is neither asynchronous nor expensive, and it avoids the race condition where `withThrowingTaskGroup` might delay.

Sources/PowerSync/Implementation/sync/SyncSignals.swift
```diff
@@ -0,0 +1,207 @@
...
+        let id: Int64
+        let continuation: AsyncThrowingStream<Void, any Error>.Continuation
+    }
+
+    /// Tracks connect-time validation for pending checkpoint request creation.
+    ///
+    /// Request ID allocation waits on this state so `next_checkpoint_request_id` only runs after
+    /// the service has affirmed or seeded the local counter for the active connection.
+    private struct PendingCheckpointRequestsState {
+        var isReady = false
```
simolus3: This sounds like a non-trivial refactoring, so treat it as optional / something to potentially revisit later: I think the architecture here would be so much simpler if this could be a reference to a `Task` we could await. Structurally this is already a deferred promise, but we know the underlying work is a seeding task we could await here.

We would create that task at the beginning of a download iteration to cater for the initial seeding and potential user-id changes, and remove it when a download iteration finishes. Requesting a checkpoint awaits the task, and if no task is currently set then we're not connecting or connected and it's illegal to request a checkpoint.

Sources/PowerSync/Implementation/CheckpointRequestImpl.swift
```diff
@@ -0,0 +1,77 @@
...
+                // `asFlow()` emits the current status first. We intentionally fail fast if the
+                // sync client is already in an error state when the caller starts waiting.
+                throw CheckpointWaitError.errorDetected(message: String(describing: error))
+            }
+
+            if !update.connected && !update.connecting {
+                throw CheckpointWaitError.disconnected
+            }
+
+            try Task.checkCancellation()
```
simolus3: This also shouldn't be necessary,  the `await` iteration should throw on cancellation.

Sources/PowerSync/Implementation/CheckpointRequestImpl.swift
stevensJourney: Discussed offline. Typed errors are difficult to work with when `CancellationErrors` need to be considered. Not implementing this.

Demos/CustomCheckpointDemo/Sources/CustomCheckpointDemo/main.swift
stevensJourney: That is a nice suggestion. This makes it much easier to open. I've implemented this.

Demos/CustomCheckpointDemo/Sources/CustomCheckpointDemo/main.swift
stevensJourney: Good catch. This is not necessary, we do sometimes need to check _after_ the consumption of `asFlow` though. But in this case, this can be and has been removed.

Demos/GRDBDemo/GRDBDemo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
stevensJourney: 👍 Updated both demos

Sources/PowerSync/Implementation/sync/Instruction.swift
stevensJourney: Removed this for now

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
stevensJourney: Yes! This is nicer! Implemented this now.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
stevensJourney: Not in this case, I can see how the flow for this was blurry. I've cleaned up this flow a bit. This is now implemented on a higher level where the intention should be clearer.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
stevensJourney: 👍 I've updated this to use a `fatalError` which should at least give more context to a crash.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
stevensJourney: I've updated the implementations to use StringEncodedInt64

Sources/PowerSync/Implementation/sync/SyncSignals.swift
stevensJourney: Yes, this is a good suggestion. I've implemented this improvement

Sources/PowerSync/Implementation/CheckpointRequestImpl.swift
stevensJourney: The in-loop check is redundant, since the next await ends the iteration anyway.

The one after the loop has to stay, though. `SyncStatus.asFlow()` returns a non-throwing `AsyncStream`, and cancelling a task suspended in `AsyncStream.next()` finishes the stream and returns `nil` rather than throwing:

So without a check the loop falls through to throw `CheckpointWaitError.disconnected`, and cancelling a pull-to-refresh reports "the sync client disconnected before the checkpoint request was synced". I've added a regression test that fails with exactly that if the check is dropped, plus a comment on the line.

Sources/PowerSync/Implementation/sync/SyncSignals.swift
stevensJourney: (This response was generated with the help of AI)

Agreed that this is structurally a deferred promise, and the lifecycle you describe is exactly what's implemented. `markPendingCheckpointRequestsRequiringAffirmation()` runs in `ActiveSyncIteration.run()`'s defer, so the state is already scoped to a download iteration. The seeding work is also already a Task (`checkpointRequestStateSeed`), so most of the pieces are there.

Three things the current machinery does that a plain task reference would need to keep, noting them mostly so whoever picks this up has them written down:

1. Waking the download loop. `waitForCheckpointRequestsReady()` defaults to `wakeDownloadLoop: true`, and the `requestCheckpoint()` path relies on it — registering a waiter cuts short the download loop's retry backoff so a pull-to-refresh while offline gets a connection attempt immediately instead of waiting out the delay. Awaiting a task doesn't produce that signal, so it'd have to stay in some form.

2. Waiting across iterations. A caller that arrives while connecting currently waits through several failed iterations until one seeds successfully. With one task per iteration, awaiting task N that failed leaves you needing task N+1, which doesn't exist yet — so you'd need a way to await "the next task", which is a deferred promise again unless it becomes something like an AsyncStream of tasks.

This also touches your "no task set means it's illegal to request" rule: during the retry backoff between iterations there's no task, but activeSync != nil, so today the request waits rather than failing. Worth deciding deliberately — failing there is defensible, it's just a behaviour change.

3. Terminal vs retryable failures. instanceNotSupported is terminal and fails pending callers via failPendingCheckpointRequests, while other seed failures leave callers waiting for the next iteration to retry. Awaiting a task rethrows whatever it failed with, so that distinction would need reintroducing at the await site.

None of these are blockers — mostly they say the replacement needs a "wait for the next successful seed" primitive rather than "await this specific task". Happy to track it as a follow-up issue rather than grow this PR further.

Sources/PowerSync/Implementation/sync/SyncSignals.swift
stevensJourney: I've updated the comments here

Tests/PowerSyncTests/SyncTests.swift
```diff
@@ -3,8 +3,109 @@ import Foundation
 @testable import PowerSync
 import Testing
 
-@Suite()
+@Suite(.serialized)
```
simolus3: Why is this necessary now, due to `minimumCheckpointRequestRetryDelayMutex`? We should at least document that, also I'm wondering if it would be possible to make `minimumCheckpointRequestRetryDelay` a debug-only instance field on `PowerSyncDatabaseImpl`? Then the service could read it from there in debug builds and use the constant in release builds. By configuring this per-instance, we wouldn't have this issue.

simolus3 (commented):
The implementation looks good to me. I didn't check the tests in detail, the suite being serialized is a minor red flag but I don't think that should necessarily delay landing this.

simolus3 (approved):
:shipit:

Tests/PowerSyncTests/SyncTests.swift
stevensJourney: Thanks for pointing this out. In development, I noticed that the tests would sometimes freeze. This was a quick fix, but, as you mentioned - it is rather suspicious. 

I managed to track the root cause of the freezing. It's not due to `minimumCheckpointRequestRetryDelayMutex`, but rather a general race condition which seemed to be more prevalent when many tests were running concurrently. 

The main issue was related to checkpoint requests, the setting of local targets and the emission of Checkpoints containing corresponding `write_checkpoint` associations in tests.

In particular. The `uploadsOfflineWrites` test was one example of this. The Mock HTTP client was always responding with a checkpoint_request_id of `1`, while the actual progression of the local `next_checkpoint_request_id` and eventual target op would increase... The main issue was that in some racy conditions, the target op would actually be set to `3`, while the emitted checkpoint from the mocked HTTP client would only contain `write_checkpoint: 2`. This would cause the local target not to get matched - which left the test handing forever. 

I've updated the Crud tests to take the actual target into account.

PR #154, opened 2026-06-30
Fix sync status observation

simolus3:
`SyncStatus.asFlow()` emits the same `SwiftSyncStatus` instance whenever it's changed (as that instance is mutable). With the existing `@State private var status: SyncStatusData` pattern we use in demos, updating the variable to itself means that sync status updates can be lost. The fix is to emit a fresh snapshot instance each time (as done with `copySnapshot()` in this PR).

Even with that fixed, the pattern of having to listen to the status flow in every SwiftUI view interested in it is not optimal. Swift has a builtin implementation of the observer pattern through the `@Observable` macro, which allows tracking updates in a more natural way. This adds the `ObservableSyncStatus` class, exposed through `db.currentStatus.observable`, to expose status updates with that mechanism.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@
   as-is so the database can live in an App Group container shared with app extensions.
   Plain filenames keep the existing behavior. The SDK coordinates opening the database to
   avoid conflicts and can share update notifications across the main app and extensions.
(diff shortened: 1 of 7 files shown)


stevensJourney (approved):
Happy with the changes. The `ObservableSyncStatus` looks very cool!

PR #159, opened 2026-07-14
Enable passing custom URL sessions

### LINKED ISSUE #158: allow custom headers on powersync connection
currently it seems to not be able to pass custom http headers to the server which i would need e.g. for cloudflare tunnels authentication. 

https://powersync-ja.github.io/powersync-swift/documentation/powersync/connectoptions

it would be good to allow passing arbitrary headers towards the sync server.

simolus3:
This allows passing a custom `URLSession` instance to `SyncClientConfiguration`. By creating a custom URL session, one can pass custom delegates observing the request lifecycle, or pass [additional headers](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/httpadditionalheaders).

I've thought about exposing a custom http client interceptor instead, but this seems like a nicely minimal API addition that still unblocks custom headers.

Closes https://github.com/powersync-ja/powersync-swift/issues/158.


stevensJourney (approved):
Happy with these changes

PR #166, opened 2026-07-22
Exchange the concrete changed tables across processes

### LINKED ISSUE #165: Self-delivered cross-process change signal causes a permanent watch-query storm with App Group databases
Fix open in PR [#166](https://github.com/powersync-ja/powersync-swift/pull/166). This issue documents the underlying problem and the reasoning behind the fix.

## Summary

When a database is opened at an absolute path (an App Group container, the documented setup for sharing with app extensions), the app process enters a permanent re-query loop: every `watch` query re-runs and re-emits identical rows every ~65-115 ms indefinitely, with zero application writes. CPU stays pinned (measured ~348% on an iPad) and the battery drains.

The same app with the database in the default sandbox directory does not exhibit this. That matches the code: `AsyncConnectionPool` only creates a `CrossProcessChangeSignal` for `.atPath` locations, so only App Group databases are affected.

## Environment

- powersync-swift 1.15.0 (introduced in #147, shipped in 1.15.0)
- iOS 26 / 27
- Database opened at an absolute App Group path, app connected and syncing
- Reproduces on device and simulator

## Evidence

The same watch query emits identical results at a fixed cadence for minutes, with the last real application write more than a minute earlier:

```
20:57:08.235 [GoalRepository] Emitting 4 active goals: ... (identical values)
20:57:08.350 [GoalRepository] Emitting 4 active goals: ...
20:57:08.477 [GoalRepository] Emitting 4 active goals: ...   <- every ~115 ms, forever
...
20:57:09.942 [StreamingSyncClient] crud upload: notify completion  <- every ~1 s, nothing to upload
```

The storm persists while the app is idle in the foreground. The `crud upload: notify completion` line fires roughly every second (the CRUD throttle) with an empty upload queue.

## Root cause

The cross-process signal is delivered back to the process that posted it, and that self-delivery is pure redundant amplification.

1. `NativeConnectionPool.write` runs `dispatchWrites` in a `defer` after every write. It calls `powersync_update_hooks('get')` and, if any tables were affected, invokes `handleUpdates`.
   - `NativeConnectionPool.swift`: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/sqlite3/NativeConnectionPool.swift#L38-L69

2. `handleUpdates` (wired in `AsyncConnectionPool.PoolOpener.obtainPool`) does two things on every local write:
   - dispatches the precise affected tables to `tableUpdates` (this is all local watchers need), and
   - calls `changeSignal.post()` to wake other processes.
   - `AsyncConnectionPool.swift`: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/AsyncConnectionPool.swift#L232-L241

3. `CrossProcessChangeSignal` posts a Darwin notification, and deliveries to the posting process itself are deliberately not suppressed (documented in the class comment). The receive handler dispatches `EXTERNAL_CHANGES_MARKER`.
   - `CrossProcessChangeSignal.swift`: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/CrossProcessChangeSignal.swift

4. `EXTERNAL_CHANGES_MARKER` is by contract treated as "potentially matching every table". It bypasses every per-table watch filter and it triggers the upload client's `ps_crud` check.
   - `watch.swift` filter: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/queries/watch.swift#L30-L34
   - `StreamingSyncClient` upload trigger: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift#L40-L42

There are two distinct problems, and the first holds even without a fully closed loop:

- Redundant amplification (always present). For every local write, the precise per-table update is already delivered via the native update hooks. The self-delivered catch-all marker adds nothing for the writing process except forcing every watcher that does NOT watch the affected tables to re-run, plus a `ps_crud` check. Cost is N watchers x M writes.

- Self-sustaining loop (the storm). Any bookkeeping commit made as a consequence of a marker wake-up re-enters the same path. `EXTERNAL_CHANGES_MARKER` -> `uploadLoop` -> `uploadAllCrud` -> `uploadLocalTarget`, which under a pending `$local` target can `UPDATE ps_buckets ...` inside a write transaction. That commit goes through `NativeConnectionPool.write` -> `dispatchWrites` -> `handleUpdates` -> `changeSignal.post()`, which the process re-delivers to itself, re-triggering the marker. The observed cadence matches the internal watch throttle (30 ms default) plus query time.
   - `uploadLocalTarget` commit path: https://github.com/powersync-ja/powersync-swift/blob/4507e96787f256264de56ce524e70ac91873f369/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift#L109-L139

An aggravating interaction: each extension process spawn (for example a widget timeline request) runs `powersync_init` / `powersync_replace_schema`, which commits and posts, so routine widget reloads wake every watcher in the app.

## Minimal reproduction

1. Open a `PowerSyncDatabase` at an absolute App Group path.
2. Register one or more `watch` queries.
3. Connect and let the client sync.
4. Observe the watch callbacks re-firing with identical rows at a fixed sub-second cadence with no application writes, and CPU pinned.

Opening the same database with a plain filename (default sandbox directory) does not reproduce it, because no `CrossProcessChangeSignal` is created for non-`.atPath` locations.

## Fix

The accompanying PR [#166](https://github.com/powersync-ja/powersync-swift/pull/166) applies both changes:

- Suppress own-process deliveries. A single sender-stamp scheme is racy: a coalesced external change could be misattributed to self and dropped, which would be a missed cross-process update. Instead each process owns a monotonic counter (one slot in a small file next to the database) and bumps only its own slot on post; a receiver fires the marker only when another process's slot advanced. Coalescing stays safe because the counters are monotonic, and nothing is lost for the posting process, since its own writes already reach its watchers via the native update hooks. It degrades to the previous always-fire behavior on any I/O error or slot exhaustion, so a change is never missed.
- Coalesce external wake-ups to one per interval, so a burst (for example rapid widget process spawns, each committing during `powersync_init`) cannot amplify into a re-query burst.

Together these break the storm and also remove the steady-state N x M redundant re-queries.

asiergmorato:
Fixes #165.

This replaces the earlier counter-file version of this PR with the local-table approach suggested in review, since it is less code and lets us propagate the exact changed tables.

## Problem

Databases opened at an absolute path (App Group container) share update notifications through a Darwin signal so writes in one process wake `watch` queries in others. That signal carries no payload, so every receiver re-emitted a catch-all `EXTERNAL_CHANGES_MARKER` that matches every table. Two consequences:

- Every process re-ran **all** of its watch queries on every write, including its own (N x M re-queries).
- It could sustain a loop. A write to table A self-delivers the catch-all, which re-runs a watch over an unrelated table B, whose handler writes A again. A device trace pinned this down: the attachment queue watches a synced table and writes its local attachments table, so with an App Group path each write woke it again roughly every 60ms and pinned the CPU (~348% measured). 998 of 1000 of those writes were immediately preceded by a self-delivered marker, so it was marker-driven, not a timer, and not the crud/upload loop. Without the shared path the same write only dispatches its own table, nothing re-fires, and it settles, which is why only App Group databases were affected.

## Approach

Processes exchange the concrete changed tables through a local, non-synced `ps_swift_updates(id, author, timestamp, tables)` table:

- Each pool picks a random `author`. Before committing a write it reads the changed tables from the update hooks and records them **in the same transaction**, then posts the Darwin signal after the commit, so a change and its record are atomic and no second, separately-contended write is needed. The record runs in a `SAVEPOINT` so a bookkeeping failure never rolls back the user's write.
- Receivers read rows newer than their watermark and written by other authors, and re-emit exactly those tables. Ignoring their own `author` is what stops a process from waking itself, and the precise tables stop unrelated watches from re-running. `EXTERNAL_CHANGES_MARKER` remains only as a fallback when precise information may be incomplete (read error, undecodable payload, the table's ids regressed, or a read gap longer than the retention window, e.g. a suspended extension).
- Rows older than the retention window are pruned. Only tables another process can act on are recorded: `ps_` is reserved for PowerSync, so anything else propagates, and of the internal tables only `ps_data__*` / `ps_data_local__*` (what `watch` matches) and `ps_crud` (what the upload loop matches) are kept. Local listeners still see every changed table.

## Notable behaviour changes / tradeoffs

- `execute(sql:parameters:)` now runs its statement in a transaction so the record commits atomically with it. Statements that cannot run in a transaction (`PRAGMA`, `VACUUM`, `ATTACH`, `DETACH`, transaction control) keep using the plain write lock. This is shared with the GRDB backend (its interceptor is `nil`, so it just gets a transactional `execute`).
- The per-write recording and the signal apply to any absolute-path database, whether or not it is actually shared, since sharing cannot be detected from the path alone.
- Raw tables must not use a `ps_` prefix, or their changes would be mistaken for internal bookkeeping and not propagate. That prefix is reserved for PowerSync; happy to add schema validation for it if you would prefer.

## Open questions

- Should this recording ultimately live in the core update hook rather than the Swift write path? In the core it would be atomic and off the write-completion thread by construction, and would cover raw `execute()` without the write-path plumbing this needed.
- Is the `ps_data__*` / `ps_data_local__*` / `ps_crud` allowlist the complete set with a cross-process consumer, or is there an internal table I should also propagate?

### DIFF
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -102,7 +102,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     }
 
     private let readState = Mutex(ReadState())
-    /// Notifications from other processes, consumed by ``startReadingUpdateLog(context:)``.
+    /// Notifications from other processes, consumed by ``startReadingUpdateLog()``.
     private let updateNotifications = BroadcastStream<Void>()
     private let updateLogReader = Mutex<Task<Void, Never>?>(nil)
 
@@ -247,21 +247,25 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     /// Reads the update log whenever another process signals, merging notifications that arrive
     /// while a read is in progress (or during the throttle that follows) into a single re-read,
     /// the same way `watch` queries throttle table updates.
-    private static func startReadingUpdateLog(context: AsyncConnectionPool) {
-        let notifications = context.updateNotifications.subscribe()
-        let task = Task { [weak context] in
+    ///
+    /// The task lives as long as the pool does and is cancelled in ``close()``. It is detached
+    /// because it belongs to the pool and not to whichever caller happened to open it: that
+    /// caller's priority, task locals and cancellation must not reach this loop.
+    private func startReadingUpdateLog() {
+        let notifications = updateNotifications.subscribe()
+        let task = Task.detached { [weak self] in
             let merged = MergeItemSequence(inner: notifications)
             do {
                 for try await _ in merged {
-                    guard let context else { return }
-                    await context.performUpdateLogRead()
-                    try await sleepForSeconds(seconds: readThrottleSeconds)
+                    guard let self else { return }
+                    await self.performUpdateLogRead()
+                    try await sleepForSeconds(seconds: Self.readThrottleSeconds)
                 }
             } catch {
                 // Cancelled while reading or waiting; nothing left to do.
             }
         }
-        context.updateLogReader.withLock { $0 = task }
+        updateLogReader.withLock { $0 = task }
     }
 
     /// Rows written by this pool are skipped, so a notification caused by our own write finds
@@ -272,7 +276,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         guard let updateLog else { return }
 
         let (since, gap) = readState.withLock { state in
-            (state.watermark, state.lastReadAt.map { Date().timeIntervalSince($0) })
+            (state.watermark, state.lastReadAt.map(Date().timeIntervalSince))
         }
         // Rows are only pruned once they are older than the retention window, so having read
         // within that window proves nothing we still need was pruned. A longer gap does not
@@ -291,11 +295,9 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
                     sql: CrossProcessUpdateLog.readSQL,
                     parameters: [.int64(since), .int64(author)],
                     callback: { rows in
-                        while let row = try rows.next(callback: { cursor -> (Int64, String) in
-                            (try cursor.getInt64(index: 0), try cursor.getString(index: 1))
-                        }) {
-                            maxSeen = row.0
-                            if let decoded = try? JSONDecoder().decode([String].self, from: Data(row.1.utf8)) {
+                        while let row = try rows.next(callback: CrossProcessUpdateLog.readRow) {
+                            maxSeen = row.id
+                            if let decoded = try? JSONDecoder().decode([String].self, from: Data(row.tables.utf8)) {
                                 tables.formUnion(decoded)
                             } else {
                                 decodeFailed = true
@@ -326,9 +328,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     }
 
     func close() async throws {
-        // Only stop listening once the pool is really closed: `close()` can be cancelled, and
-        // the database stays usable in that case, so it must keep receiving other processes'
-        // changes. A notification arriving while we close is harmless, the read just fails.
+        // Only stop update notifications after the async close, since that can be cancelled.
         try await self.opener.close()
         updateLog?.signal.stop()
         updateLogReader.withLock { reader in
@@ -357,7 +357,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
             // re-enter this actor and build a second pool.
             context.readState.withLock { $0.lastReadAt = Date() }
             if context.updateLog != nil {
-                AsyncConnectionPool.startReadingUpdateLog(context: context)
+                context.startReadingUpdateLog()
                 context.updateLog?.signal.start { [weak context] in
                     context?.updateNotifications.dispatch(event: ())
                 }
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -75,17 +75,41 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     private let logger: any LoggerProtocol
     private let tableUpdatesStream = BroadcastStream<Set<String>>()
     private let opener = PoolOpener()
-    /// Cross-process change signaling; `nil` for in-memory databases (nothing to share).
-    private let changeSignal: CrossProcessChangeSignal?
+    /// Set for databases other processes can open, in which case writes are announced through
+    /// ``CrossProcessUpdateLog`` and this pool reacts to the other processes' rows.
+    ///
+    /// Note that this covers any absolute path, not just App Group containers, since sharing
+    /// cannot be detected from the path alone.
+    private let updateLog: CrossProcessUpdateLog?
 
     init(location: DatabaseLocation, logger: any LoggerProtocol, initialStatements: [String] = []) {
         self.location = location
         self.logger = logger
         self.initialStatements = initialStatements
-        self.changeSignal = location.sharedPath.map {
-            CrossProcessChangeSignal(databasePath: $0, logger: logger)
+        self.updateLog = location.sharedPath.map {
+            CrossProcessUpdateLog(
+                signal: CrossProcessChangeSignal(databasePath: $0, logger: logger),
+                logger: logger
+            )
         }
     }
+
+    /// Tracks the highest consumed update-log row id and when the log was last read, so a
+    /// long read gap (a suspended process) can be distinguished from steady-state reads.
+    private struct ReadState {
+        var watermark: Int64 = 0
+        var lastReadAt: Date?
+    }
+
+    private let readState = Mutex(ReadState())
+    /// Notifications from other processes, consumed by ``startReadingUpdateLog(context:)``.
+    private let updateNotifications = BroadcastStream<Void>()
+    private let updateLogReader = Mutex<Task<Void, Never>?>(nil)
+
+    /// Minimum spacing between update-log reads. Notifications arriving while we read or wait
+    /// are merged into a single further read, so a burst of writes cannot become a burst of
+    /// reads. Cross-process latency is not critical: a process sees its own writes immediately.
+    private static let readThrottleSeconds: TimeInterval = 0.25
 
     var tableUpdates: AsyncStream<Set<String>> {
         tableUpdatesStream.subscribe()
@@ -128,6 +152,12 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
             }
 
             let _ = try context.execute(sql: "select powersync_update_hooks('install')", parameters: [])
+
+            // Create the update log before any write can try to record into it (the
+            // `powersync_init()` write during initialization is the first one).
+            if updateLog != nil {
+                let _ = try context.execute(sql: CrossProcessUpdateLog.createTableSQL, parameters: [])
+            }
         }
     }
 
@@ -152,7 +182,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         try configureConnection(connection: writer, isWriter: true)
 
         if case .inMemory = location {
-            return NativeConnectionPool(singleConnection: writer, logger: logger, handleUpdates: handleUpdates)
+            return NativeConnectionPool(singleConnection: writer, logger: logger, updateLog: updateLog, handleUpdates: handleUpdates)
         }
         let numReaders = 4
         var readers = RigidDeque<RawSqliteConnection>(capacity: numReaders)
@@ -161,7 +191,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
             try configureConnection(connection: reader, isWriter: false)
             readers.append(reader)
         }
-        return NativeConnectionPool(writer: writer, readers: readers, logger: logger, handleUpdates: handleUpdates)
+        return NativeConnectionPool(writer: writer, readers: readers, logger: logger, updateLog: updateLog, handleUpdates: handleUpdates)
     }
 
     /// Builds the pool, retrying with asynchronous backoff while another process holds the
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         }
     }
 
+    /// Reads the update log whenever another process signals, merging notifications that arrive
+    /// while a read is in progress (or during the throttle that follows) into a single re-read,
+    /// the same way `watch` queries throttle table updates.
+    private static func startReadingUpdateLog(context: AsyncConnectionPool) {
+        let notifications = context.updateNotifications.subscribe()
+        let task = Task { [weak context] in
+            let merged = MergeItemSequence(inner: notifications)
+            do {
+                for try await _ in merged {
+                    guard let context else { return }
+                    await context.performUpdateLogRead()
+                    try await sleepForSeconds(seconds: readThrottleSeconds)
+                }
+            } catch {
+                // Cancelled while reading or waiting; nothing left to do.
+            }
+        }
+        context.updateLogReader.withLock { $0 = task }
+    }
+
+    /// Rows written by this pool are skipped, so a notification caused by our own write finds
+    /// nothing to do. Falls back to ``EXTERNAL_CHANGES_MARKER``, which re-runs every watch, when
+    /// the precise tables may be incomplete: a read error, an undecodable payload, or a read gap
+    /// long enough that rows we needed could have been pruned (a suspended process).
+    private func performUpdateLogRead() async {
+        guard let updateLog else { return }
+
+        let (since, gap) = readState.withLock { state in
+            (state.watermark, state.lastReadAt.map { Date().timeIntervalSince($0) })
+        }
+        // Rows are only pruned once they are older than the retention window, so having read
+        // within that window proves nothing we still need was pruned. A longer gap does not
+        // mean we lost anything (an idle database prunes nothing either, since pruning happens
+        // on writes), but we cannot tell, so we re-query everything once and then go back to
+        // precise updates.
+        let mightHaveMissed = gap.map { $0 > Double(CrossProcessUpdateLog.retentionSeconds) } ?? true
+
+        do {
+            let author = updateLog.author
+            let result: (tables: Set<String>, maxSeen: Int64?, decodeFailed: Bool) = try await read { connection in
+                var tables = Set<String>()
+                var maxSeen: Int64?
+                var decodeFailed = false
+                try connection.withIterator(
+                    sql: CrossProcessUpdateLog.readSQL,
+                    parameters: [.int64(since), .int64(author)],
+                    callback: { rows in
+                        while let row = try rows.next(callback: { cursor -> (Int64, String) in
+                            (try cursor.getInt64(index: 0), try cursor.getString(index: 1))
+                        }) {
+                            maxSeen = row.0
+                            if let decoded = try? JSONDecoder().decode([String].self, from: Data(row.1.utf8)) {
+                                tables.formUnion(decoded)
+                            } else {
+                                decodeFailed = true
+                            }
+                        }
+                    }
+                )
+                return (tables, maxSeen, decodeFailed)
+            }
+
+            readState.withLock { state in
+                if let maxSeen = result.maxSeen, maxSeen > state.watermark {
+                    state.watermark = maxSeen
+                }
+                state.lastReadAt = Date()
+            }
+
+            if mightHaveMissed || result.decodeFailed {
+                tableUpdatesStream.dispatch(event: [EXTERNAL_CHANGES_MARKER])
+            } else if !result.tables.isEmpty {
+                tableUpdatesStream.dispatch(event: result.tables)
+            }
+        } catch {
+            // Reading failed (including when the pool is closing), so we cannot tell which
+            // tables changed: conservatively re-query everything.
+            tableUpdatesStream.dispatch(event: [EXTERNAL_CHANGES_MARKER])
+        }
+    }
+
     func close() async throws {
-        changeSignal?.stop()
+        // Only stop listening once the pool is really closed: `close()` can be cancelled, and
+        // the database stays usable in that case, so it must keep receiving other processes'
+        // changes. A notification arriving while we close is harmless, the read just fails.
         try await self.opener.close()
+        updateLog?.signal.stop()
+        updateLogReader.withLock { reader in
+            reader?.cancel()
+            reader = nil
+        }
     }
 
     private actor PoolOpener {
@@ -230,17 +348,21 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
             try registerPowerSyncCoreExtension()
             let handleUpdates: @Sendable (_: Set<String>) -> () = { [weak context] updates in
                 context?.tableUpdatesStream.dispatch(event: updates)
-                // Tell other processes sharing this file that tables changed.
-                context?.changeSignal?.post()
-            }
-            context.changeSignal?.start { [weak context] in
-                // Another process (or this one; harmless, throttled downstream) changed
-                // the database outside this pool's update hooks.
-                context?.tableUpdatesStream.dispatch(event: [EXTERNAL_CHANGES_MARKER])
             }
 
             let pool = try await context.buildPoolWithRetry(handleUpdates: handleUpdates)
             self.pool = pool
+
+            // Listen only once the pool exists, so a notification arriving while we open cannot
+            // re-enter this actor and build a second pool.
+            context.readState.withLock { $0.lastReadAt = Date() }
+            if context.updateLog != nil {
+                AsyncConnectionPool.startReadingUpdateLog(context: context)
+                context.updateLog?.signal.start { [weak context] in
+                    context?.updateNotifications.dispatch(event: ())
+                }
+            }
+
             return pool
         }
 
--- a/Sources/PowerSync/Implementation/CrossProcessChangeSignal.swift
+++ b/Sources/PowerSync/Implementation/CrossProcessChangeSignal.swift
@@ -1,20 +1,20 @@
 import Foundation
 import notify
 
-/// Cross-process change signaling for a database file, built on Darwin notifications —
-/// the same mechanism Core Data uses for its remote-change notifications.
+/// Cross-process change transport for a database file, built on Darwin notifications, the
+/// same mechanism Core Data uses for its remote-change notifications.
 ///
-/// PowerSync's update hooks only observe writes made through the local connection pool.
-/// When several processes share a database file (an app and its widgets or App Intents
-/// extensions), each process posts this signal after committing a write; the others
-/// re-emit their `tableUpdates` with ``EXTERNAL_CHANGES_MARKER`` so `watch` queries
-/// re-run and the upload client checks `ps_crud`.
+/// PowerSync's update hooks only observe writes made through the local connection pool. When
+/// several processes share a database file (an app and its widgets or App Intents
+/// extensions), each process records the tables it changed in the `ps_swift_updates` table
+/// and posts this signal; the others read the new rows and re-emit those tables on their
+/// `tableUpdates` so `watch` queries re-run and the upload client checks `ps_crud`.
 ///
 /// Darwin notifications carry no payload and are coalesced under pressure, which is fine:
-/// the marker means "something changed, re-query". Deliveries to the posting process
-/// itself are not suppressed — a redundant re-query (already throttled) is preferable to
-/// the race a sender-stamp suppression scheme introduces, where an external change could
-/// be misattributed and silently dropped.
+/// the payload lives in `ps_swift_updates`, and each receiver tracks the highest row id it
+/// has consumed. Deliveries to the posting process itself are not suppressed here either; that
+/// is handled where the rows are read, by ignoring this process's own `author`, so a process
+/// never wakes itself.
 final class CrossProcessChangeSignal: @unchecked Sendable {
     private let name: String
     private let logger: any LoggerProtocol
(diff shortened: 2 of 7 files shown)


asiergmorato:
### Code review

Self-review of the initial commit (`7fe2382`) surfaced 4 issues, all addressed in `c75e457`.

1. `SharedProcessCounters.close()` was not idempotent, risking a write through a reused file descriptor. `fileDescriptor` was a `let` left `>= 0` after `Foundation.close`, and `stop()` reaches `close()` from both the explicit `AsyncConnectionPool.close()` and `deinit`, so the second call could `flock`/`pwrite` on a descriptor the OS had reassigned to an unrelated file.

https://github.com/powersync-ja/powersync-swift/blob/7fe2382d038a2dfec72f79c235d22377298ccd50/Sources/PowerSync/Implementation/SharedProcessCounters.swift#L103-L112

Fixed: descriptor reset to `-1` and slot cleared on close, second call is a no-op.

2. Blocking file I/O on the (typically main) thread. The counter table was opened, `flock`ed, and probed with up to 64 `kill()` calls inside the synchronous `init` chain reached from the synchronous `PowerSyncDatabase(...)` initializer, so a peer holding the lock (a suspended widget) could stall app launch.

https://github.com/powersync-ja/powersync-swift/blob/7fe2382d038a2dfec72f79c235d22377298ccd50/Sources/PowerSync/Implementation/SharedProcessCounters.swift#L43-L72

Fixed: moved to `prepare()`, called from `start()` on the background pool-open path; the baseline is still captured before the listener registers.

3. Silent fallback. Falling back to no-suppression (open failure or slot exhaustion) had no log, unlike the sibling Darwin-registration failure in `start()`, so a return of the storm would be undiagnosable. Fixed by passing the logger in and warning on both fallback paths.

4. Doc over-claimed "a change is never missed". Self-suppression assumes every writer participates in the table; a non-participating writer (fallback mode, or an older SDK version during an app-update window) does not bump a slot, so a suppressing sibling can miss its change until the next observed signal.

https://github.com/powersync-ja/powersync-swift/blob/7fe2382d038a2dfec72f79c235d22377298ccd50/Sources/PowerSync/Implementation/SharedProcessCounters.swift#L19-L21

Fixed: documented the participation assumption honestly instead of over-claiming.

Also considered and dismissed: a PID-reuse slot leak (degrades gracefully to the pre-PR behavior) and a create/delete path canonicalization mismatch for the sidecar file (both paths resolve to the same inode, so no leak).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

simolus3 (commented):
Thanks for following up on this! To reduce the amount of code needed for this feature, I wonder if we could use a local SQLite table instead of a second shared file.

Potentially, we could have a `CREATE TABLE ps_swift_updates(author INTEGER, timestamp INTEGER, tables TEXT)` on the database. Each PowerSync database generates a random value for `author` and inserts into `ps_swift_updates` before posting a Darwin notification. When receiving a notification, a process collects updated tables from other authors and keeps track of the highest `rowid` it has seen in that table to avoid ever seeing a notification twice. To prevent that table from growing too much, we could delete entries older than a few seconds before writing.

But I need to understand the actual loop in more detail first, since from my understanding even a `uploadCrud` loop should settle eventually when it runs out of things to do.

asiergmorato:
You're right that uploadCrud settles on its own once ps_crud is empty and the $local target is resolved. The perpetuation doesn't come from crud piling up. It comes from the self-delivered catch-all marker re-triggering the upload machinery:

1. uploadLoop fires on ps_crud or EXTERNAL_CHANGES_MARKER.
2. Every iteration calls notifyCrudUploadComplete()  → watchCompletedCrudUploads() dispatches localEvents(.completedUpload).
3. localEvents is merged into controlArgs, so .completedUpload runs powersyncControl(...), which is a db.writeTransaction. That commit registers affected ps_ tables via the update hooks.
4. dispatchWrites → handleUpdates → changeSignal.post() → the same process receives its own Darwin notification → EXTERNAL_CHANGES_MARKER → back to step 1.

The powersyncControl write only touches system ps_ tables, which no user watch matches, so user watches are woken solely by the catch-all marker, and that marker only exists because the post is delivered back to the sender. Remove the self-delivery and the loop opens: the process's own bookkeeping write no longer comes back as a catch-all that re-arms its own uploadLoop, so uploadLoop is driven only by real ps_crud changes and genuine external markers, and it settles. crudThrottle bounds the crud-completion cadence to ~1s (matching your crud upload: notify completion log); the faster watch cadence lines up with the extra precise writes and the second (widget) process, but the closure above is the necessary enabler in every case.

On the SQLite table side, it may be worth considering. It removes the second file and the whole mmap/flock/liveness primitive, and it lets us do something the current design cannot: carry the actual changed tables across processes.

simolus3:
Is there a realistic test / repro in this PR that fails without these changes? I'm still surprised about this behavior and it sounds like there's something else worth fixing to break the loop.

>localEvents is merged into controlArgs, so .completedUpload runs powersyncControl(...), which is a db.writeTransaction. That commit registers affected ps_ tables via the update hooks.

It only does that if there's something to write, specifically a pending checkpoint that we couldn't apply before due to outstanding local data (we'd try again after `.completedUpload`). So that call should either do nothing, or apply a checkpoint which then triggers one more iteration of the notify loop but not more.

So if you have a way to reproduce this, I'd love to step through that with a debugger to understand what's responsible for the continued writes.

asiergmorato:
**Repro / what's actually driving the loop**

Got a device trace with per-process logging. You were right that it's not the crud / `completedUpload` loop (that shows up only ~68 times in a minute and doesn't perpetuate). The perpetuating write is the **attachment queue writing its local table** (`ps_data_local__route_attachments`), and the loop is closed by the self-delivered catch-all marker.

The cycle, repeating every ~60 ms, entirely in the main app:

```
[COMMIT] tables=["ps_data_local__route_attachments"]
[POST] -> notify_post -> [SIGNAL recv]   (self-delivery, same process, same ms)
[RECV] -> EXTERNAL_CHANGES_MARKER
   -> re-runs every watch, including the AttachmentQueue's watchAttachments
   -> the queue writes route_attachments again
[COMMIT] tables=["ps_data_local__route_attachments"]   (repeat)
```

998 of 1000 `route_attachments` commits are immediately preceded by a self-delivered marker, so it's marker-driven, not a timer.

**Why it only happens with an App Group path:** `watchAttachments` watches a synced table, while the queue writes the local `route_attachments` table. Single-process, that write dispatches precisely `["ps_data_local__route_attachments"]`, which doesn't match the attachment watch, so nothing re-fires. With the App Group path the same write self-delivers the catch-all `EXTERNAL_CHANGES_MARKER`, which matches every table, re-fires `watchAttachments`, and the queue writes again. So it isn't one component spinning; it's the catch-all marker crossing table boundaries, connecting a writer of table A to a watcher of an unrelated table B whose handler writes A. (It also re-runs unrelated watches, like our goals query, on every cycle.)

**The `ps_swift_updates` approach**

We prototyped your suggestion and it fixes this well in the real app: https://github.com/Chubby-Studio/powersync-swift/tree/experiment/ps-swift-updates-log

- Receivers read rows `id > watermark AND author != self` and re-emit the **precise** tables, so `EXTERNAL_CHANGES_MARKER` becomes only a gap/failure fallback. A process never wakes itself (the `author` filter replaces the counter file), and a `route_attachments` write never wakes a watch over a different table, so both the loop and the N x M amplification go away.
- The row is written **inside the same transaction** as the write (a before-commit hook on the write path), so it commits atomically with the data: no lost-change window and no separate contendible write on the completion thread.

**One question for you**

The trace also shows a lot of internal-only writes during normal sync (`ps_buckets`, `ps_oplog`, `ps_tx`, `ps_stream_subscriptions`, ...), each currently recording a row and posting a signal for no cross-process consumer (watches match user tables, the upload loop matches `ps_crud`). We could skip recording/posting when the changed tables are purely internal bookkeeping, but that needs the authoritative list of internal tables that never have a cross-process consumer (a prefix allowlist won't work because of raw tables). Which tables would you consider safe to never propagate cross-process?

If you prefer this direction, we'll switch this PR over to it.

simolus3:
> but that needs the authoritative list of internal tables that never have a cross-process consumer (a prefix allowlist won't work because of raw tables).

I actually think a prefix is likely fine if you make it a denylist, skip `ps_` tables that aren't `ps_data__` or `ps_data_local__`-prefixed or `ps_crud`. Raw tables should not have a `ps_` prefix, that can reasonably be reserved for us.

Even though it adds a bit more complexity in lines of code from a quick look, I think the benefits of being able to track exact tables and not having to roll our own cross-process scheme kind of make me lean towards the approach based on the local table.

asiergmorato:
Updated this PR to the local-table approach we discussed (force-pushed over the counter-file version, now a single commit). It exchanges the concrete changed tables through a local `ps_swift_updates(id, author, timestamp, tables)` table:

- Each pool has a random `author`; the changed tables are recorded inside the same transaction as the write, and the Darwin signal is posted after commit, so the record is atomic with the write.
- Receivers read rows newer than their watermark from other authors and re-emit the precise tables, so a process never wakes itself and unrelated watches don't re-run. `EXTERNAL_CHANGES_MARKER` stays only as a fallback (read error, undecodable payload, id regression, or a read gap past the retention window).
- Applied your prefix suggestion: only `ps_data__*` / `ps_data_local__*` / `ps_crud` and non-`ps_` tables are recorded/signalled, so internal sync bookkeeping (`ps_buckets`, `ps_oplog`, ...) no longer generates cross-process chatter. Local listeners still get everything.

Full write-up is in the updated description. Two things I would value your take on:

1. Whether this recording should ultimately live in the core update hook instead of the Swift write path (it would be atomic and off the write-completion thread by construction, and cover raw `execute()` without the write-path plumbing).
2. Whether `ps_data__*` / `ps_data_local__*` / `ps_crud` is the complete set of internal tables with a cross-process consumer.

The diagnostic trace that pinned the loop (attachment queue + self-delivered catch-all) is summarized in the description under Problem.

asiergmorato:
Hey @simolus3, hope you're doing well! Just gently circling back on this one, no rush at all, I know you have a lot on your plate.
If there's anything that would make the review easier, I'm happy to help: more tests, splitting it into smaller pieces, or a quick walkthrough of the write-path change.

Sources/PowerSync/Implementation/sqlite3/NativeConnectionPool.swift
```diff
@@ -132,6 +134,17 @@ struct RawSqliteConnection: ~Copyable {
         precondition(!closed)
         return NativeConnectionLease(pointer: self.connection)
     }
+
+    /// Like ``asLease()`` but throws instead of trapping when the connection has been closed.
+    /// The pool can hand back a just-closed connection to a waiter that raced ``close()``
+    /// (for example a cross-process signal read landing during teardown); throwing turns that
+    /// into a catchable error rather than a crash.
+    func leaseOrThrow() throws -> NativeConnectionLease {
```
simolus3: We'll fix this in a separate PR: https://github.com/powersync-ja/powersync-swift/pull/177

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -70,21 +70,137 @@ enum DatabaseLocation {
...
+        self.changeSignal = signal
+        self.commitInterceptor = Self.makeCommitInterceptor(
+            signal: signal, author: author, stream: tableUpdatesStream, logger: logger
+        )
+    }
+
+    /// Tracks the highest consumed update-log row id and when the log was last read, so a
+    /// long read gap (a suspended process) can be distinguished from steady-state reads.
+    private final class ReadState: @unchecked Sendable {
+        private let lock = NSLock()
```
simolus3: Please use our `Mutex` helper for that.

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -128,6 +244,11 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
             }
 
             let _ = try context.execute(sql: "select powersync_update_hooks('install')", parameters: [])
+
+            // Create the cross-process update log before any write can try to record into it
+            // (e.g. the `powersync_init()` write during initialization). Recreated again after
+            // each `powersync_replace_schema` in case the core reconciliation drops it.
```
simolus3: > Recreated again after each `powersync_replace_schema` in case the core reconciliation drops it.

It won't

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,7 +335,230 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+    /// deadlock). Also prunes stale rows to keep the table small. Returns whether the row was
+    /// written, so the caller only wakes other processes when there is something to read.
+    fileprivate func recordUpdateLog(tables: Set<String>, lease: NativeConnectionLease) -> Bool {
+        // Bound cross-process write-lock waiting for this best-effort bookkeeping: it runs
+        // synchronously on the write-completion path (a cooperative-pool thread), so without a
+        // cap it could stall that thread for the connection's full 30s busy_timeout under
+        // contention. Under sustained contention the record is skipped rather than blocking.
+        // (Moving the whole write-completion dispatch off the cooperative pool is a deeper,
+        // separate change.)
+        let _ = try? lease.execute(sql: "PRAGMA busy_timeout = 250", parameters: [])
```
simolus3: We run this after a write has already happened, which could block just as long. So I don't think this is necessary, but if we want this then it should at least run a `pragma busy_timeout` to restore it to the actual value instead of the default after.

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,7 +335,230 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+        logger: any LoggerProtocol
+    ) -> WriteCommitInterceptor? {
+        guard let signal else { return nil }
+        let table = updateLogTable
+        let retention = updateLogRetentionSeconds
+        let insertSQL = insertUpdateLogSQL
+        let pruneSQL = pruneUpdateLogSQL
+
+        return WriteCommitInterceptor(
+            recordBeforeCommit: { context in
```
simolus3: I don't really understand the purpose of this indirection, why can't we run these queries directly in `NativeConnectionPool.dispatchWrites`?

simolus3 (commented):
Sorry for the slow response here. I did not look at the changes in detail yet, but I think the overall architecture is sound. I've added a few comments, mainly I think we should simplify this by potentially passing the state (when cross-process updates are enabled) directly to `NativeConnectionPool` which would then manage the update records.

asiergmorato:
Thanks @simolus3, and no worries about the delay!

I've reworked it along those lines, it's a good bit simpler now:

- The cross-process state is now a CrossProcessUpdateLog handed to NativeConnectionPool, which appends the changed tables in dispatchWrites and posts the notification. That removes the interceptor indirection entirely, along with the transaction plumbing it needed, so TransactionImpl, QueriesProtocol and LeaseContext are untouched now.
- Rebased onto main so the lease fix from #177 is used instead of mine.
- Switched the reader state to the Mutex helper.
- Dropped the defensive re-creation after powersync_replace_schema and, with it, the id-regression handling that only existed for that case. The read is a single query now.
- Removed the busy_timeout juggling: with the record happening on the write path there is no second contended write to bound.

One tradeoff worth flagging: recording after the commit (rather than inside the transaction) means a crash in that narrow window loses the notification for other processes, and the write path does the insert on the write-completion thread. Happy to revisit if that ever matters, but it keeps this much simpler and matches what you had in mind.

Full test suite passes locally.

asiergmorato:
Hi @simolus3! Do you have any estimate on when this is going to be shipped? We currently are using the repo anchored to this commit because otherwise the app really goes hard on resources otherwise. Thanks for your help.

asiergmorato:
CI failed on requestCheckpointWaitsUntilApplied, which looks unrelated: it's an in-memory database (so the cross-process path is inactive), the assertion is about sync status right after disconnect(), and the same job passed that test on one destination and failed it on the other. It passes locally 10/10, and the full suite passes. Happy to look closer if you've seen it flake before.

Sources/PowerSync/Implementation/CrossProcessChangeSignal.swift
```diff
@@ -47,7 +62,27 @@ final class CrossProcessChangeSignal: @unchecked Sendable {
         }
     }
 
-    /// Posts the signal; called after every committed write.
+    /// Emits `onChange` immediately if the last emission is older than ``coalesceInterval``,
+    /// otherwise schedules a single trailing emission at the end of the current window.
+    /// Must run on `queue`.
+    private func dispatchCoalesced(onChange: @escaping @Sendable () -> Void) {
```
simolus3: Instead of debouncing here, is it possible to post into an `AsyncStream` that we throttle via a `MergeItemSequence` and a task adding a delay when listening? That might simplify this a bit, ignore if it turns out to make things more complicated.

asiergmorato:
Hi @simolus3! Good call, that simplified it more than expected: MergeItemSequence already does the merging I had hand-rolled, so both the debounce here and a read gate I had in the pool are gone. The signal is now just the transport, and a task consumes the notifications through MergeItemSequence with a throttle after each read, the same shape watch() uses.

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         }
     }
 
+    /// Reads the update log whenever another process signals, merging notifications that arrive
+    /// while a read is in progress (or during the throttle that follows) into a single re-read,
+    /// the same way `watch` queries throttle table updates.
+    private static func startReadingUpdateLog(context: AsyncConnectionPool) {
+        let notifications = context.updateNotifications.subscribe()
+        let task = Task { [weak context] in
```
simolus3: Should this be a detached task? We cancel the task on `close()`, and cancelling the context that originally called `obtainPool` should not auto-cancel this task. So I believe opting out of structured concurrency would be better here, or at least add a small comment about the intended lifecycle of this task.

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         }
     }
 
+    /// Reads the update log whenever another process signals, merging notifications that arrive
+    /// while a read is in progress (or during the throttle that follows) into a single re-read,
+    /// the same way `watch` queries throttle table updates.
+    private static func startReadingUpdateLog(context: AsyncConnectionPool) {
```
simolus3: Given that this is defined in `AsyncConnectionPool`, why not make it a `fileprivate` instance method instead of a static function?

Sources/PowerSync/Implementation/sqlite3/NativeConnectionPool.swift
```diff
@@ -43,8 +50,15 @@ final class NativeConnectionPool: Sendable {
                     return try decoder.decode(Set<String>.self, from: try $0.getString(index: 0).data(using: .utf8)!)
                 }
 
-                if let affectedTables, !affectedTables.isEmpty {
-                    self.handleUpdates(affectedTables)
+                // Our own writes to the update log must not feed back into the machinery.
+                let changed = (affectedTables ?? []).subtracting([CrossProcessUpdateLog.tableName])
```
simolus3: `subtracting` might copy, can we declare `affectedTables` as a mutable variable and use `remove()` instead?

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+
+    /// Rows written by this pool are skipped, so a notification caused by our own write finds
+    /// nothing to do. Falls back to ``EXTERNAL_CHANGES_MARKER``, which re-runs every watch, when
+    /// the precise tables may be incomplete: a read error, an undecodable payload, or a read gap
+    /// long enough that rows we needed could have been pruned (a suspended process).
+    private func performUpdateLogRead() async {
+        guard let updateLog else { return }
+
+        let (since, gap) = readState.withLock { state in
+            (state.watermark, state.lastReadAt.map { Date().timeIntervalSince($0) })
```
simolus3: nit:
```suggestion
            (state.watermark, state.lastReadAt.map(Date().timeIntervalSince))
```

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+        let mightHaveMissed = gap.map { $0 > Double(CrossProcessUpdateLog.retentionSeconds) } ?? true
+
+        do {
+            let author = updateLog.author
+            let result: (tables: Set<String>, maxSeen: Int64?, decodeFailed: Bool) = try await read { connection in
+                var tables = Set<String>()
+                var maxSeen: Int64?
+                var decodeFailed = false
+                try connection.withIterator(
+                    sql: CrossProcessUpdateLog.readSQL,
```
simolus3, line 295: Let's keep the definition of the SQL statement text and the code reading from cursors close to each other. Like this, I need to switch between files to check whether the statement/read combo is correct.

We could either inline the constant here, or add a static function to `CrossProcessUpdateLog` to read a `(Int64, String)` from `any SqlCursor`.

Sources/PowerSync/Implementation/AsyncConnectionPool.swift
```diff
@@ -214,9 +244,97 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
...
+            // tables changed: conservatively re-query everything.
+            tableUpdatesStream.dispatch(event: [EXTERNAL_CHANGES_MARKER])
+        }
+    }
+
     func close() async throws {
-        changeSignal?.stop()
+        // Only stop listening once the pool is really closed: `close()` can be cancelled, and
+        // the database stays usable in that case, so it must keep receiving other processes'
+        // changes. A notification arriving while we close is harmless, the read just fails.
```
simolus3: nit: Please shorten this comment, maybe just say something like "Only stop update notifications after async close since that can be cancelled".

Sources/PowerSync/Implementation/CrossProcessUpdateLog.swift
```diff
@@ -0,0 +1,90 @@
...
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            author INTEGER NOT NULL,
+            timestamp INTEGER NOT NULL,
+            tables TEXT NOT NULL
+        )
+        """
+
+    /// Rows older than this are pruned on each write. A receiver that falls behind by more than
+    /// this window can no longer rely on the rows it missed still being there, so it detects the
+    /// gap and re-queries everything instead.
```
simolus3: Maybe be more specific and point to how that works:

```suggestion
    /// Rows older than this are pruned on each write. A receiver that falls behind by more than
    /// this window can no longer rely on the rows it missed still being there, so it emits a
    /// generic ``EXTERNAL_CHANGES_MARKER`` instead.
```

Sources/PowerSync/Implementation/CrossProcessUpdateLog.swift
```diff
@@ -0,0 +1,90 @@
...
+    /// Appends the tables changed by a write, on the connection that just wrote them, and prunes
+    /// stale rows. Returns whether a row was recorded, so the caller only signals other processes
+    /// when there is something for them to read. Never throws: failing to record must not fail
+    /// the write that already happened.
+    func record(tables: Set<String>, lease: some SQLiteConnectionLease) -> Bool {
+        let relevant = tables.filter(Self.isRelevant)
+        guard !relevant.isEmpty else { return false }
+
+        do {
+            let json = String(decoding: try JSONEncoder().encode(relevant.sorted()), as: UTF8.self)
```
simolus3: Why do we need to sort them here, we don't seem to rely on that?

Tests/PowerSyncTests/Implementation/CrossProcessPreciseTests.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    /// A process must not announce its own writes twice: they are dispatched locally when the
+    /// write commits, and the notification it receives back for them must find nothing, since
+    /// rows are read with `author != self`. Without that filter every write would be announced a
+    /// second time, which is what fed the re-query loop.
+    ///
+    /// Counting on `tableUpdates` rather than through `watch`, because a watch query coalesces
+    /// two updates arriving close together into a single re-run and would hide the difference.
+    @available(iOS 16, macOS 13, watchOS 9, tvOS 16, *)
+    @Test(.timeLimit(.minutes(1)))
```
simolus3: The timeout here feels very generous, can we remove it? It looks like we expect this test to complete in roughly 1.5s after all. Same for the other test.

simolus3 (commented):
The approach and implementation here looks good to me, I only have a few minor comments remaining.

asiergmorato:
All addressed, in commits on top of what you reviewed.

- **Detached task:** switched to `Task.detached` and documented the lifecycle. One correction on the reasoning: an unstructured `Task {}` is already not cancelled by the task that created it (only structured children are), so cancellation was never leaking in. It did inherit priority and task locals from whichever caller happened to open the pool, which is wrong for a loop that lives as long as the pool, so detached is the better fit either way.
- **Instance method:** `startReadingUpdateLog()` is now a private instance method.
- **`subtracting`:** the set is bound as a `var` straight from the cursor and the log table is dropped with `remove()`, so the usual case no longer allocates a second set.
- **Cursor read:** added `CrossProcessUpdateLog.readRow(_:)` right next to `readSQL`, so the statement and the column reads stay together.
- Applied the doc suggestion on `retentionSeconds`, shortened the `close()` comment, took the `lastReadAt` nit, and dropped the `sorted()` (nothing depended on the order).
- **Test time limits:** removed. That also made the `@available(iOS 16, macOS 13, ...)` on those tests unnecessary, since `.timeLimit` was the only thing that needed it, so it is gone too (checked that the test bundle still builds for the iOS Simulator).

Full suite passes, including with `-strict-concurrency=complete`.

simolus3 (approved):
Thanks for the contribution! This looks good to me.

PR #168, opened 2026-07-23
Refactor http clients to expand test coverage

simolus3:
We currently use a `HttpClient` protocol exposing `/sync/stream` lines as strings. This allows mocking clients in tests, giving us good coverage of the higher-level sync client logic. However, the real implementation for `HttpClient` remains untestable.

Recently, we found an issue in the way we convert byte streams to lines that can't be covered by tests. To make more parts of the stack testable, this replaces the `HttpClient` protocol with a lower-level `PowerSyncUrlSession` protocol that only covers `URLSession.bytes()` and `URLSession.data()`. Everything else (reading lines from byte streams, logging, throwing on unknown content types) can now use a mocked `URLSession`, allowing us to test much more of the HTTP client stack.

The diff in `HttpClient.swift` is a bit tricky to follow, but basically the changes are:

1. Add the `PowerSyncUrlSession` protocol which is essentially a wrapper around `URLSession`.
2. Remove `HttpClient` as a protocol as well as the `PlatformHttpClient` and `LoggingClient` implementations.
3. Re-introduce client functionality as `SpecializedHttpClient`, using Swift generics to specialize to a particular `PowerSyncUrlSession` implementation.
   - A logger is passed to those methods, replacing the need for a wrapping logging client implementation. 
   - Move the bytes -> lines mapping to `SpecializedSyncLineResponseIterator`, the implementation is basically the same.
5. Add type-erased wrappers (`SyncLineResponseIterator`, `SyncLineResponse` and `HttpClient`) as concrete structs used throughout the SDK.

For performance reasons, an `HttpClient` implementation is specialized with its underlying session implementation. This is important: `URLSession.bytes` returns an `AsyncSequence<UInt8>`, giving us an async method returning a single byte at the time. For acceptable performance, we need the compiler to be able to inline the `next()` call to `URL.AsyncBytes`. Doing that without adding generics around http client throughout the SDK is surprisingly tricky, Claude suggested using closures to erase the type without despecializing. This works, but is kind of ugly. Other recommendations are definitely welcome!

This also adds a check to throw when a streaming HTTP response ends without a trailing newline. This should never happen with the PowerSync service, but we have similar checks in other SDKs so it's useful to have it here as well.

AI use: I used Claude to validate that we get a specialized implementation in `readLine` and let it make some refactorings towards that. Tests are also AI-generated, but reviewed manually.


stevensJourney (approved):
Happy with the changes.

> For performance reasons, an HttpClient implementation is specialized with its underlying session implementation. This is important: URLSession.bytes returns an AsyncSequence<UInt8>, giving us an async method returning a single byte at the time. For acceptable performance, we need the compiler to be able to inline the next() call to URL.AsyncBytes. Doing that without adding generics around http client throughout the SDK is surprisingly tricky, Claude suggested using closures to erase the type without despecializing. This works, but is kind of ugly. Other recommendations are definitely welcome!

I think the current solution is a good compromise :)

PR #169, opened 2026-07-24
Simplify release process

simolus3:
To create releases for the Swift SDK, we use a manually-triggered GitHub action taking the version and release notes as input:

grafik

This is kind of clumsy to use, and we run the action for no reason at all. The only thing our release workflow does is create a tag and push it, there are no attached assets or any extra build steps. GitHub has a page providing the exact same thing, why aren't we using that instead?

grafik

I assume we originally added that action to anticipate a custom build step, but in its current state it's just a somewhat annoying step. This removes the workflow and simplifies the release guide for subsequent releases.


Chriztiaan (approved):
Makes sense to me. Thanks for updating the release doc too!

PR #171, opened 2026-07-30
Fix race condition in MergeItemSequence dropping errors

simolus3:
An error state in `MergeItemSequence` should be final, as it can only happen in a `catch` block after the task broke out of its loop. Currently, a `defer` block above the do/catch would transition the state to `.done` shortly after reporting the error, dropping the intermediate error state.

Instead of representing events as results, this adds a dedicated error state that transitions to `.done` when observed, removing the need for the broken `defer`.

AI use: Changes are manual, issue was originally found with Codex after a CI failure on https://github.com/powersync-ja/powersync-swift/pull/153. Claude found a few more potential issues around a cancelled call to `next()` potentially dropping errors, but this should make the test more reliable and in SDK uses we don't actually let errors flow into `MergeItemSequence`.

### DIFF
--- a/Sources/PowerSync/Utils/MergeItemSequence.swift
+++ b/Sources/PowerSync/Utils/MergeItemSequence.swift
@@ -29,14 +29,14 @@ struct MergeItemSequence<Base: AsyncSequence & Sendable>: AsyncSequence where Ba
         init(inner: Base) {
             let state = IteratorState()
             self.pollTask = Task {
-                defer { state.inner.withLock { $0.transitionToDone() } }
-
                 do {
-                    for try await event in inner {
-                        state.inner.withLock { $0.markHasEvent(event: .success(event)) }
+                    for try await _ in inner {
+                        state.inner.withLock { $0.markHasEvent() }
                     }
+
+                    state.inner.withLock { $0.transitionToDone() }
                 } catch {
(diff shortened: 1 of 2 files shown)


Sources/PowerSync/Utils/MergeItemSequence.swift
```diff
@@ -74,7 +74,12 @@ private enum MergeSequenceState {
     /// We're waiting in next() for an upstream emission.
     case waitingForUpstream(CheckedContinuation<()?, any Error>)
     /// We have an upstream emission that has not yet been sent (due to backpressure or throttle).
-    case hasPendingEvent(Result<(), any Error>)
+    case hasPendingEvent
+    /// Fetching from upstream failed.
+    /// 
+    /// This is a final state, we just need to ensure it gets emitted downstream before transitioning
```
stevensJourney: Nit: The wording here is a bit confusing at first read. It feels like `failure` is temporary and `done` is the only `final` state here.

stevensJourney (dismissed):
Changes look good :)

Sources/PowerSync/Utils/MergeItemSequence.swift
simolus3: I have expanded the comment a bit: `failure` is technically temporary after all, it gets cleared after a call to `next()`.

stevensJourney (approved):

PR #175, opened 2026-08-01
Fix issue with weak let in Sendable AirDatabaseGroup

will-lundy:
Fix: Correct weak var declaration and Sendable conformance in ActiveDatabaseGroup

Summary
This PR fixes two compiler issues in ActiveDatabaseGroup related to weak reference declarations and Sendable conformance.

Changes
weak let → weak var

weak references must be declared as var in Swift, since they are implicitly mutable — the runtime sets them to nil when the referenced object is deallocated. The previous weak let declaration was invalid Swift and would not compile.

Sendable → @unchecked Sendable

The compiler cannot automatically verify Sendable conformance on ActiveDatabaseGroup due to the presence of the weak var reference. Marking the class as @unchecked Sendable resolves the compiler error and is appropriate here given the existing design of the type.

⚠️ Note: @unchecked Sendable shifts thread-safety responsibility to the developer. Concurrent access patterns on ActiveDatabaseGroup should be reviewed to ensure this conformance is genuinely safe.

### DIFF
--- a/Sources/PowerSync/Implementation/ActiveInstanceStore.swift
+++ b/Sources/PowerSync/Implementation/ActiveInstanceStore.swift
@@ -44,9 +44,9 @@ private final class ActiveDatabaseGroupData: Sendable {
 /// Additionally, we want to avoid two databases in the same group having a sync stream open at the same time to avoid
 /// duplicate resources being used. For this reason, each active database group has a single sync coordinator actor
 /// responsible for initializing the sync process for all databases in the group.
-final class ActiveDatabaseGroup: Sendable {
+final class ActiveDatabaseGroup: @unchecked Sendable {
     fileprivate let data: ActiveDatabaseGroupData
-    private weak let collection: DatabaseGroupCollection?
+    private weak var collection: DatabaseGroupCollection?
 
     fileprivate init(data: ActiveDatabaseGroupData, collection: DatabaseGroupCollection) {
         self.data = data
--- a/Sources/PowerSync/Implementation/ActiveInstanceStore.swift
+++ b/Sources/PowerSync/Implementation/ActiveInstanceStore.swift
@@ -45,6 +45,7 @@ private final class ActiveDatabaseGroupData: Sendable {
 /// duplicate resources being used. For this reason, each active database group has a single sync coordinator actor
 /// responsible for initializing the sync process for all databases in the group.
 final class ActiveDatabaseGroup: @unchecked Sendable {
+// unchecked sendable because we need to use weak var (our minimum Swift version is 6.1, weak let was introduced in 6.3).
     fileprivate let data: ActiveDatabaseGroupData
     private weak var collection: DatabaseGroupCollection?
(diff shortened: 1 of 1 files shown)


simolus3 (approved):
> weak references must be declared as var in Swift, since they are implicitly mutable — the runtime sets them to nil when the referenced object is deallocated. The previous weak let declaration was invalid Swift and would not compile.

Which compiler options did you use? I agree this needs to be a `weak var`, but it compiles without emitting a warning. Maybe there's a flag we're missing that would allow us to catch this in our CI?

will-lundy:
> > weak references must be declared as var in Swift, since they are implicitly mutable — the runtime sets them to nil when the referenced object is deallocated. The previous weak let declaration was invalid Swift and would not compile.
> 
> Which compiler options did you use? I agree this needs to be a `weak var`, but it compiles without emitting a warning. Maybe there's a flag we're missing that would allow us to catch this in our CI?

We were upgrading to Xcode 26.6.  I think the 26.6 compiler is more strict.  The new compiler rules caught this and our CI is failing on this line.

Sources/PowerSync/Implementation/ActiveInstanceStore.swift
```diff
@@ -44,9 +44,9 @@ private final class ActiveDatabaseGroupData: Sendable {
 /// Additionally, we want to avoid two databases in the same group having a sync stream open at the same time to avoid
 /// duplicate resources being used. For this reason, each active database group has a single sync coordinator actor
 /// responsible for initializing the sync process for all databases in the group.
-final class ActiveDatabaseGroup: Sendable {
+final class ActiveDatabaseGroup: @unchecked Sendable {
```
simolus3, line 47: Could you add a comment explaining this as well?

```suggestion
final class ActiveDatabaseGroup: @unchecked Sendable {
// unchecked sendable because we need to use weak var (our minimum Swift version is 6.1, weak let was introduced in 6.3).
```

simolus3 (dismissed):
Are you sure it's the upgraded versions? I now see that `weak let` is a thing as of [Swift 6.3](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0481-weak-let.md), and it's designed for this exact scenario. So I wonder if it's actually an older XCode version that might be causing trouble here.

Either way, we declare a minimum version of Swift 6.1 so we shouldn't rely on this.

Sources/PowerSync/Implementation/ActiveInstanceStore.swift
will-lundy: Done

will-lundy:
So, we were specifying the version of Xcode on Bitrise, not the version of swift.  Its possible that it is falling back to 6.1 for some reason.  Locally, this is compiling fine with XCode 26.5.  It also might be down to the difference in the actual machine that is compiling this on bitrise.  Not sure.

simolus3 (approved):

PR #176, opened 2026-08-03
Update core extension

simolus3:
This updates the core extension to 0.5.2, fixing a memory leak when some functions like `powersync_control('target_checkpoint_request_id')` are called on a read-only connection. This is not an issue for the Swift SDK which only calls that in write transactions, but it's still good to stay up-to-date.

This also attempts to fix a few tests that regularly broke CI:

- `disconnectCancelsDefaultCheckpointRequestRetryDelay()` compared the delay to 1 second, which a slow CI runner can exceed. Increase to compare against the actual limit of 5 seconds.
- `checkpointRequestRetryWaitsAgainAfterNewRequest` used a short retry delay of .2 seconds, which also breaks in slow CI runners. Increase to 2s.


stevensJourney (approved):
Thanks for the fixes!

Issue #172, opened 2026-07-30
NativeConnectionPool has no terminal closed state: close() returns closed connections to the semaphore, so a queued read/write traps on precondition(!closed) in asLease()


amitmai:
**SDK version:** 1.15.0 (rev `4507e96`), also present on `main` as far as I can tell. Reproduced on iOS/macOS test hosts.

## Symptom

```
PowerSync/NativeConnectionPool.swift:132: Precondition failed
```

A hard process trap (not a throw) during database teardown. In our case it kills the whole test process; in an app it would be a crash on sign-out / account switch.

Frequently preceded in the log by the *safe* branch of the same race on a sibling operation:

```
Attempted to use closed PowerSync database
```

## Mechanism

`NativeConnectionPool.close()` acquires the writer and every reader, then closes the `RawSqliteConnection`s **in place**:

```swift
func close() async throws {
    var write = try await writer.acquire(count: 1)
    var acquiredReaders: SemaphoreGrant<RawSqliteConnection>? = nil
    if let readers { acquiredReaders = try await readers.acquire(count: readers.count) }
    write.acquiredItems[0].close()
    // ... readers closed similarly
}
```

When those `SemaphoreGrant`s go out of scope, `SemaphoreGrant.deinit` calls `semaphore.returnItems(items:)` (`Utils/AsyncMutex.swift`) — so the **now-closed** connections are returned to the semaphore's available pool. The pool itself keeps no closed state, so the next acquirer (including one already queued behind `close()`) receives a closed connection and calls:

```swift
func asLease() -> NativeConnectionLease {
    precondition(!closed)   // <- line 132
    ...
}
```

The only guard is on a different isolation domain: `DatabaseInitializationAction.ensureInitialized` in `PowerSyncDatabaseImpl.swift`, which throws `"Attempted to use closed PowerSync database"`. But `close()` flips that actor's `closed` flag and *then* runs `pool.close()`, and every operation checks the flag before suspending on `pool.read`/`pool.write`. So:

1. Operation A calls `initialize()` → passes the `closed == false` check.
2. A suspends in `semaphore.acquire`.
3. `close()` flips `closed = true`, closes the connections, grants deinit → dead connections returned.
4. A acquires a dead connection → `asLease()` → **trap**.

Check-then-acquire with an unbounded suspension in between.

## Why this is hard to defend against as a caller

[...]

PR #177, opened 2026-08-03
Fix concurrency issues around `close()`

simolus3:
This fixes two issues reported in https://github.com/powersync-ja/powersync-swift/issues/172.

### Crashes on race between `close()` and other operations

Emitting a closed SQLite connection as a connection lease crashes the process (as a `precondition`). This is better than the use-after-free it would have been otherwise, but it shouldn't be possible to reach that state. Unfortunately, it can happen when:

1. We call `close()`.
2. Another task runs an operation on the database. Because `close()` did not complete yet, `DatabaseInitializationAction.ensureInitialized()` works and the operation can continue. It waits for a connection, though.
3. The original task closes connections and returns, returning closed connections into the async semaphore.
4. The second task now obtains the closed connection, crashing.

The fix is what we do on most other SDKs too: Check for closed connections _after_ we've obtained their handle from an async semaphore.

### Cancelling `close()`

When a call to `close()` is cancelled before it actually closes connections, it would previously leave the database in a state where everything throws even though connections are still open.

This fixes the issue by only marking the database as closed when connections are actually closed. There's still a race condition here where while a database is being closed, queries would throw but if the throw is cancelled the database keeps working afterwards. I think that should be harmless overall, though.

AI use: Original issue report is likely generated by AI. Changes are manual, with minor revisions after a review with Claude Code.

### DIFF
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -221,7 +221,6 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
 
     private actor PoolOpener {
         private var pool: NativeConnectionPool? = nil
-        private var isClosed = false
 
         func obtainPool(pool context: AsyncConnectionPool) async throws -> NativeConnectionPool {
             if let pool {
@@ -246,11 +245,6 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
         }
 
         func close() async throws {
-            if isClosed {
-                return
-            }
-
-            isClosed = true
             if let pool {
                 try await pool.close()
             }
--- a/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
+++ b/Sources/PowerSync/Implementation/AsyncConnectionPool.swift
@@ -101,7 +101,7 @@ final class AsyncConnectionPool: SQLiteConnectionPoolProtocol {
     }
 
     private func configureConnection(connection: borrowing RawSqliteConnection, isWriter: Bool) throws {
-        let context = connection.asLease()
+        let context = try connection.asLease()
         for stmt in initialStatements {
             let _ = try context.execute(sql: stmt, parameters: [])
         }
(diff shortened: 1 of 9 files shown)


Sources/PowerSync/Implementation/AsyncConnectionPool.swift
stevensJourney: I was chatting about the PR with Codex (about a different question). In that chat Codex pointed out that there may be an issue if the `PowerSyncDatabase` is `closed`, cancelled and closed again. The second close might keep the pool's connections open - due to skipping the last close. 

Here's a small test which fails.

```swift
    @Test func canCloseAfterCancellingClose() async throws {
        let pool = AsyncConnectionPool(
            location: .inMemory,
            logger: DefaultLogger()
        )
        let db = OpenedPowerSyncDatabase(
            schema: Schema(),
            pool: pool,
            identifier: "cancel-then-close-test"
        )

        let (readerAcquired, signalReaderAcquired) = AsyncStream<Void>.makeStream()
        let releaseReader = DispatchSemaphore(value: 0)
        let reader = Task {
            try await db.readLock { _ in
                signalReaderAcquired.yield()
                releaseReader.wait()
            }
        }

        // Wait until the callback has leased a reader, then initiate close independently.
        var readerAcquiredIterator = readerAcquired.makeAsyncIterator()
        _ = await readerAcquiredIterator.next()
        let cancelledClose = Task { try await db.close() }
        cancelledClose.cancel()
        await #expect(throws: CancellationError.self) {
            try await cancelledClose.value
        }
        releaseReader.signal()
        try await reader.value

        // Retrying close must still reach and close the underlying connection pool.
        try await db.close()

        // We inspect the pool directly because the database's initialization actor now considers
        // the database closed and would reject public operations before they reach the pool. A
        // successful retry closes every RawSqliteConnection, so acquiring a lease must throw. If
        // this succeeds, PoolOpener kept isClosed = true after the cancelled attempt and treated
        // the retry as a no-op, leaving the native connections open.
        await #expect(throws: PowerSyncError.self) {
            try await pool.read { _ in () }
        }
    }
```

It seems like we might want to revert the `isClosed` flag here on cancellation also?

simolus3: Nice catch, I've added that test. I don't think `PoolOpener` needs a dedicated closed state since the underlying `NativePool` prevents concurrency around close, it's harmless to call its `close()` method multiple times and we have the database-level `DatabaseInitializationAction` as a guard.

stevensJourney (approved):
:shipit:

PR #183, opened 2026-09-09
Apply subscription changes made while the sync stream is being established

### LINKED ISSUE #157: ~20s stall when re-fetching data after TTL eviction of a sync stream subscription (connection stays connected=true, no data delivered)
## Environment

- **Client:** powersync-swift 1.14.4 (also reproduced on 1.14.3), iOS app, HTTP sync stream
- **PowerSync Service:** 1.23.2, self-hosted (Docker via Coolify), MongoDB bucket storage
- **Source DB:** Postgres (self-hosted Supabase), Supabase JWT auth
- **Sync config:** Sync Streams, edition 3, **no subscription parameters used** (`params: [:]`)

## Context

I am testing the Sync Stream TTL mechanism in a small counter test app before using it
in a production app. For this test app alone, `auto_subscribe: true` would be perfectly
fine (small dataset, always needed). But in the production app I want on-demand
subscriptions with TTL so that stale local data gets cleaned up automatically and is
re-fetched from the backend when needed again.

The TTL eviction itself works as expected. The problem is what happens **after** the
eviction: once the local cache has been cleared, it takes **~20 seconds (sometimes more,
sometimes less)** until the data is re-downloaded from the server into the local SQLite
database — during which the connection reports `connected=true` but nothing is delivered.

**Important:** This does NOT affect the normal `connect()` path. Initial sync with
`auto_subscribe: true` is very fast, and subscribing *before* `connect()` is also fast.
The stall occurs specifically in the TTL flow: subscription expired → bucket cleared
locally → re-subscribe on the live connection → long wait for the re-fetch.

## Stream definition

```yaml
streams:
  counters:
    # no auto_subscribe → on-demand, client subscribes explicitly
    query: SELECT * FROM public.counters WHERE user_id = auth.user_id()

  users:
    auto_subscribe: true
    query: SELECT * FROM public.users WHERE id = auth.user_id()
```

## Client code

Subscriptions are managed via a small wrapper. The default TTL is 7 days; for testing
I set it to 60 seconds (also tested with 5 minutes — same behavior; with 7 days it
would be the same, just after 7 days):

```swift
import Foundation
import PowerSync

@Observable
@MainActor
final class StreamSubscriptionManager {
    static let shared = StreamSubscriptionManager()

    nonisolated static let defaultTTL: TimeInterval = 7 * 24 * 60 * 60  // 7 days; tests override via the ttl parameter (e.g. 60s)

    @ObservationIgnored
    private var subscriptions: [String: any SyncStreamSubscription] = [:]

    private var database: (any PowerSyncDatabaseProtocol)? {
        PowerSyncStore.shared.database
    }

    private init() {}

    func subscribe(
        _ streamName: String,
        ttl: TimeInterval = StreamSubscriptionManager.defaultTTL
    ) async {
        guard subscriptions[streamName] == nil else { return }
        guard let database else { return }
        do {
            subscriptions[streamName] = try await database
                .syncStream(name: streamName, params: [:])
                .subscribe(ttl: ttl, priority: nil)
        } catch {
            print("Could not subscribe to \(streamName): \(error.localizedDescription)")
        }
    }

    func unsubscribe(_ streamName: String) async {
        if let sub = subscriptions[streamName] {
            try? await sub.unsubscribe()
            subscriptions.removeValue(forKey: streamName)
        }
    }

    func unsubscribeAll() async {
        for (_, sub) in subscriptions {
            try? await sub.unsubscribe()
        }
        subscriptions.removeAll()
    }
}
```

Connecting follows the standard pattern from the docs (Supabase connector providing
endpoint + access token via `fetchCredentials`):

```swift
func connect() async {
    guard let database else { return }
    do {
        try await database.connect(connector: connector)
    } catch {
        self.error = error
        print("Could not connect: \(error.localizedDescription)")
    }
}
```

App flow: `connect()` runs first, then `subscribe("counters", ttl: 60)`. On logout,
`unsubscribeAll()` is called (which starts the TTL), followed by `disconnect()`.

## Reproduction

1. Log in, subscribe to `counters` with `ttl: 60`. Data syncs and displays normally.
2. Log out (`unsubscribeAll()` + `disconnect()`) — the TTL starts.
3. Wait > 60 seconds.
4. Log back in **or** fully restart the app (killed, not backgrounded) and let it
   connect + subscribe again.
5. Shortly after connecting, the expired bucket is cleared locally (expected —
   the local table drops to 0 rows).
6. **The re-fetch then stalls for ~20 seconds** before the data arrives from the
   server. During the stall, sync status reports `connected=true`,
   `downloading=false`. Then the client appears to reconnect internally, and the
   data arrives immediately after.

This happens consistently on both paths (fresh login after TTL expiry, and full app
restart after TTL expiry).

## Logs

Notes on the log lines:
- Lines starting with `🟣 / 🔵 / 🔌 [Counters] / [Status]` are **my own debug prints**
  (a SQLite watch query on the `counters` table, and a 1s poller logging every change
  of `database.currentStatus`).
- Lines starting with `[StreamingSyncClient]` come **from the PowerSync SDK** itself.

```
🟣 [Counters] emission #2 @ 18:58:51 — 0 rows → cache cleared mid-watch (expected TTL eviction)
[StreamingSyncClient] Validated and applied checkpoint
[StreamingSyncClient] crud upload: notify completion
🔌✅ [Status] @ 18:58:51 — connected=true connecting=false downloading=false hasSynced=true

        ← ~19 seconds, connection stays connected=true, nothing is delivered →

🔵 [Counters] ✅ emission #3 @ 18:59:10 — 21 rows → data re-downloaded from server and stored locally
[StreamingSyncClient] crud upload: notify completion
[StreamingSyncClient] Validated and applied checkpoint
```

An earlier run of the same repro additionally captured the status flip right before
the data arrived, showing the internal reconnect:

```
18:28:50  emission — 0 rows (bucket cleared)
18:28:51  connected=true, connecting=false, downloading=false
          ← ~20s, nothing received →
18:29:11  connected=false, connecting=true      ← spontaneous internal reconnect
18:29:11  emission — 21 rows (data arrives instantly after the reconnect)
18:29:12  connected=true
```

## What I ruled out

- **Bucket operation history / download size:** Ran the service `compact` command
  (removed 231 stale ops from the bucket) and defragmented via no-op row touches.
  No change.
- **Service version / reverse proxy buffering:** Upgraded the service 1.21.0 → 1.23.2
  (which includes the `X-Accel-Buffering` fix for `/sync/stream`). No change.
- **`retryDelay`:** Setting `retryDelay: 1` on `connect()` made no difference — the
  connection never enters a retry state during the stall (status stays
  `connected=true` the whole time).
- **Server-side checkpoint idleness:** Writing a row to Postgres during the stall
  does not unblock it.
- **Control cases:** With `auto_subscribe: true`, or when calling `subscribe()`
  *before* `connect()`, everything is instant. The stall only occurs when the
  active subscription set changes on a live connection (which is inherent to the
  TTL-eviction → re-subscribe flow).

## Suspicion

The behavior matches issues that were fixed in the sibling SDKs but may not be
ported to Swift yet:

- Dart: *"Fix changes in active Sync Stream subscriptions causing a reconnect
  delay"* — powersync.dart PR #410 (closes #409): the sync client now breaks out
  of the iteration when the core extension requests a close, instead of posting
  `AbortCurrentIteration` and racing while new streams attach.
- Kotlin: reconnect-delay fix for prefetched tokens (PR #343), same underlying
  pattern of a stop command causing the iteration to pause and wait out a delay.

It looks like the Swift sync client gets stuck in the old iteration after the
subscription set changes, until an internal reconnect (visible in the logs above)
forces a new `/sync/stream` request — at which point the data arrives immediately.

Happy to provide more logs (including service-side logs) or test a fix. Thanks!

asiergmorato:
Fixes #157

## Problem

A sync stream subscription added after `connect()` returned but before the first `/sync/stream` response arrived only took effect on the next keep-alive line from the service, typically 20 seconds later. With `includeDefaultStreams: false` and a fresh database, the common "connect, then subscribe" pattern hits this on every first launch: the first request goes out with no subscriptions at all, the service answers with an empty checkpoint (`buckets: 0`), and nothing is delivered until the keep-alive triggers the periodic subscription check in the core.

This is the same stall @elcode100 describes in #157, and the mechanism @simolus3 pointed at there: the `DidUpdateSubscriptions` notification that should have caused an immediate reconnect never reached the core.

## Cause

In `SyncIteration.run()`, `watchSyncStreams` and `watchCompletedCrudUploads` start before the request goes out, but the merge only subscribed to `localEvents` after `fetchSyncLines` returned. `BroadcastStream.dispatch` drops events that have no listener, so an `updateSubscriptions` raised in that window (a few milliseconds locally, up to the full round trip on a slow link) was lost.

## Fix

Subscribe to `localEvents` before the watchers start. The `AsyncStream` is unbounded, so anything dispatched while the request is in flight is buffered and consumed as soon as the control loop starts; the core then compares the current subscriptions against the in-flight request and reconnects right away, exactly as it already does once connected. No changes to the core extension or the protocol.

## Test

`subscriptionAddedWhileConnectingReconnects` holds the first mock response open, subscribes in that window, and requires a second `/sync/stream` request carrying the subscription without ever sending a keep-alive. It fails on 1.16.1 (times out after 5 s) and passes with the change in about 70 ms. The full package suite passes.

## Verified against a real service

PowerSync service 1.24.0 (self-hosted), iOS app, fresh anonymous account, subscribing four streams right after `connect()`:

- Before: first stream opened with `buckets: 0` and stayed idle for 19.93 s; the service logged `Sync stream complete` right after the keep-alive, then a second stream with `buckets: 16` delivered the data.
- After: first stream still opens with `buckets: 0`, `Sync stream complete` after 70 ms, second stream with `buckets: 16` 300 ms later. First row visible to the app 0.59 s after sign-in instead of 20 s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,8 @@
 
 ## Unreleased
 
-* Fix sync stream subscriptions changed while a connection was being established (after
-  `connect()` returned and before the first `/sync/stream` response arrived) taking effect only
-  on the next keep-alive line from the service, typically 20 seconds later. The change
-  notification was dispatched before the sync iteration started listening for local events, so
-  it was dropped and the core extension never learned about it until its periodic check.
+* Fix sync stream subscriptions changed while the connection was still being established only
+  taking effect on the next keep-alive line from the service, typically 20 seconds later.
 
 ## 1.16.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+* Fix sync stream subscriptions changed while a connection was being established (after
+  `connect()` returned and before the first `/sync/stream` response arrived) taking effect only
+  on the next keep-alive line from the service, typically 20 seconds later. The change
+  notification was dispatched before the sync iteration started listening for local events, so
+  it was dropped and the core extension never learned about it until its periodic check.
+
 ## 1.16.1
 
 * Fix a re-query storm on databases opened at an absolute path (App Group container). The
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 1.16.2
 
 * Fix sync stream subscriptions changed while the connection was still being established only
   taking effect on the next keep-alive line from the service, typically 20 seconds later.
--- a/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
+++ b/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
@@ -514,13 +514,10 @@ private struct ActiveSyncIteration: Sendable {
             signals.markPendingCheckpointRequestsRequiringAffirmation()
         }
 
-        // Listen for local events BEFORE the watchers below start dispatching them. The stream
-        // buffers whatever is dispatched until the control loop consumes it, so a subscription
-        // change or a completed upload that lands while the request is still being established
-        // takes effect as soon as the response arrives. Subscribing only after `fetchSyncLines`
-        // returned left those events without a listener (`BroadcastStream.dispatch` drops them),
-        // and the core extension then only noticed the change on the next keep-alive line,
-        // typically 20 seconds later.
+        // Subscribe to local events before the watchers below start dispatching them: `BroadcastStream`
+        // only delivers to listeners that already exist, and the subscription buffers everything until
+        // the control loop drains it once the sync stream is established. Otherwise a subscription
+        // change made while the request is in flight is only picked up on the next keep-alive line.
         let pendingLocalEvents = localEvents.subscribe()
 
         // Notify the core extension for changed Sync Stream subscriptions, as we might have to reconnect.
--- a/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
+++ b/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
@@ -514,10 +514,6 @@ private struct ActiveSyncIteration: Sendable {
             signals.markPendingCheckpointRequestsRequiringAffirmation()
         }
 
-        // Subscribe to local events before the watchers below start dispatching them: `BroadcastStream`
-        // only delivers to listeners that already exist, and the subscription buffers everything until
-        // the control loop drains it once the sync stream is established. Otherwise a subscription
-        // change made while the request is in flight is only picked up on the next keep-alive line.
         let pendingLocalEvents = localEvents.subscribe()
 
         // Notify the core extension for changed Sync Stream subscriptions, as we might have to reconnect.
--- a/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
+++ b/Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
@@ -514,6 +514,15 @@ private struct ActiveSyncIteration: Sendable {
             signals.markPendingCheckpointRequestsRequiringAffirmation()
         }
 
+        // Listen for local events BEFORE the watchers below start dispatching them. The stream
+        // buffers whatever is dispatched until the control loop consumes it, so a subscription
+        // change or a completed upload that lands while the request is still being established
+        // takes effect as soon as the response arrives. Subscribing only after `fetchSyncLines`
+        // returned left those events without a listener (`BroadcastStream.dispatch` drops them),
+        // and the core extension then only noticed the change on the next keep-alive line,
+        // typically 20 seconds later.
+        let pendingLocalEvents = localEvents.subscribe()
+
         // Notify the core extension for changed Sync Stream subscriptions, as we might have to reconnect.
         let (currentStreams, streamChanges) = syncClient.db.group.syncCoordinator.streams.observeActiveStreams()
         async let _ = watchSyncStreams(changes: streamChanges)
@@ -557,7 +566,7 @@ private struct ActiveSyncIteration: Sendable {
                     controlArgs = AsyncAlgorithms.merge(
                         serviceEvents,
                         checkpointRequestStateValidationEvents(task: checkpointRequestStateSeed),
-                        localEvents.subscribe()
+                        pendingLocalEvents
                     )
                 } catch {
                     let streamError = error
--- a/Tests/PowerSyncTests/SyncTests.swift
+++ b/Tests/PowerSyncTests/SyncTests.swift
@@ -2193,11 +2193,9 @@ class InMemorySyncIntegrationTests {
     }
 
     @Test func subscriptionAddedWhileConnectingReconnects() async throws {
-        // A subscription added after `connect()` returned but before the first `/sync/stream`
-        // response arrived used to be applied only on the next keep-alive line from the service,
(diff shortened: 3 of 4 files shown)


Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
```diff
@@ -514,6 +514,12 @@ private struct ActiveSyncIteration: Sendable {
             signals.markPendingCheckpointRequestsRequiringAffirmation()
         }
 
+        // Subscribe to local events before the watchers below start dispatching them: `BroadcastStream`
+        // only delivers to listeners that already exist, and the subscription buffers everything until
+        // the control loop drains it once the sync stream is established. Otherwise a subscription
+        // change made while the request is in flight is only picked up on the next keep-alive line.
```
simolus3: IMO this is obvious enough in hindsight that it doesn't need a comment :)

Tests/PowerSyncTests/SyncTests.swift
```diff
@@ -2192,6 +2192,57 @@ class InMemorySyncIntegrationTests {
...
+            }
+            if count == 1 {
+                await firstRequestStarted.complete()
+                await releaseFirstResponse.await()
+            }
+            return AsyncThrowingChannel<Data, any Error>()
+        })
+
+        // A short retry delay keeps the 5 s polling budget below independent of how the reconnect
+        // is scheduled.
```
simolus3: I don't understand why this is necessary, the reconnect because of sync stream changes shouldn't hit the retry delay right? Does the test really take 5s without it?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
```
simolus3: Could you update this to `1.16.2` (and also `CurrentVersion.swift`)? I'm not aware of other changes coming up, so we can release this after merging.

Sources/PowerSync/Implementation/sync/StreamingSyncClient.swift
asiergmorato: Gone, agreed.

Tests/PowerSyncTests/SyncTests.swift
asiergmorato: You're right, it isn't necessary and I've removed it. The test runs in about 60 ms either way: the close a subscription change produces carries `hide_disconnect: true`, so `downloadLoop` skips `waitForRetryDelayOrPendingCheckpointRequest` entirely and the reconnect never goes through the retry path.

I added it out of caution because `waitUntilAsync`'s budget is 100 x 50 ms, which is exactly the 5 s default retry delay, so the two would collide if a reconnect ever did take that path. That is hypothetical today, so it is better left out.

CHANGELOG.md
asiergmorato: Done: `1.16.2` in both `CHANGELOG.md` and `CurrentVersion.swift`.

simolus3 (approved):
Thanks for the contribution!

asiergmorato:
Hi @simolus3! Any ETA for the release of 1.16.2 with the fix?

simolus3:
Thanks for the reminder, I forgot to tag the release after merging this. I have just tagged 1.16.2.

